PR #25, opened 2016-12-06
Added isnotin method to Term

x8lucas8x:


pypika/terms.py
```diff
@@ -340,19 +345,23 @@ def __init__(self, term, container, alias=None):
             The term to assert membership for within the container.
         :param container:
             A list or subquery.
+        :param isin:
+            A boolean to alter the behaviour to not in if false.
         """
         super(ContainsCriterion, self).__init__(alias)
         self.term = term
         self.container = container
+        self.isin = isin
```
twheys: It would probably make sense to pull this up to the base class Criterion since any sort of criterion could be negated

pypika/tests/test_criterions.py
```diff
@@ -333,6 +332,45 @@ def test__function_isin(self):
         self.assertEqual('COALESCE("isin"."foo",0) IN (0,1)', str(c2))
 
 
+class IsNotInTests(unittest.TestCase):
+    t = Table('abc', alias='isnotin')
```
twheys: Could you please change the alias to something like `test_table`?  It's not immediately clear in the test assertions and gave me the impression at first of something not being serialized correctly.

pypika/terms.py
```diff
@@ -84,8 +84,13 @@ def between(self, lower, upper):
 
     def isin(self, arg):
         if isinstance(arg, (list, tuple, set)):
-            return ContainsCriterion(self, ListField([self._wrap(value) for value in arg]))
-        return ContainsCriterion(self, arg)
+            return ContainsCriterion(self, ListField([self._wrap(value) for value in arg]), isin=True)
+        return ContainsCriterion(self, arg, isin=True)
+
+    def isnotin(self, arg):
```
twheys: Could you add an override for `__not__` in criterion which sets the negation boolean, currently named `isin`?  Also, more of a nitpick, I think this function name would sound nicer as just `notin`.

x8lucas8x: @twheys There is no __not__() method for object instances; only the interpreter core defines this operation (see https://docs.python.org/3/library/operator.html?highlight=__not__). I even tried, but custom __not__ methods are never executed.

pypika/tests/test_criterions.py
x8lucas8x: @twheys Do you want me to change the alias of the other criterion tests as well? They are usually named after the criterion that is being tested.

x8lucas8x:
@twheys Please review this again.

pypika/terms.py
```diff
@@ -268,6 +271,9 @@ def get_sql(self, **kwargs):
 
 
 class Criterion(Term):
+    def __init__(self, alias=None):
+        super(Criterion, self).__init__(alias)
+        self._isNegated = False
```
twheys: Please use PEP 8 naming conventions https://www.python.org/dev/peps/pep-0008/#descriptive-naming-styles

pypika/terms.py
```diff
@@ -278,6 +284,10 @@ def __or__(self, other):
     def __xor__(self, other):
         return ComplexCriterion(Boolean.xor_, self, other)
 
+    def negate(self):
+        self._isNegated = True
```
twheys: This should be immutable, could you please annotate it with `@builder`

twheys:
Merged and published 0.1.8

PR #38, opened 2017-04-21
Added support for joins with USING and updated the documentation

### LINKED ISSUE #36: additional Join functionality
- It would be a nice convenience function just to specify a string or list of strs if the column(s) to join on are the same in both tables. e.g. from_(t1).join(t2).on('c1')

- it's not clear how I can construct a join USING clause "... from t1 join t2 USING (c1, c2)"

i'm happy to contribute if this is not yet available but you think would be helpful

I'm sorry the docs don't show a preferred method of getting in touch. please suggest most appropriate communication channel. You can look me up as "CJ Lee" from London on linkedin

P.S. love this package thanks for this

twheys:
Fixes #36 

Added support for USING joins with tests. Rewrote the join documentation to omit low level details and to cover both join types.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -289,11 +289,13 @@ After adding a ``GROUP BY`` clause to a query, the ``HAVING`` clause becomes ava
 Joining Tables and Subqueries
 """""""""""""""""""""""""""""
 
-Tables and subqueries can be joined to any query using the ``Query.join()`` method.  When joining tables and
-subqueries, a criterion must provided containing an equality between a field from the primary table or joined tables and
-a field from the joining table.  When calling ``Query.join()`` with a table, a ``TablerJoiner`` will be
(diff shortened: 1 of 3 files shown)


twheys:
@cjstudioz I would prefer to make the changes to the kwargs in a separate PR.  Not sure which format you're referring to, I can't see any that could just use the class dict as params for the format function.

cjstudioz:
Cool. I only mentioned it as a minor thing as a suggestion. As for
formatting self.__dict__ it was a more general comment since I've seen the
option to do that in many places in the code , maybe not so obvious with
only the changes in this review (sorry)

twheys:
No worries, some of the format code is admittedly a bit complex. I'd welcome any suggestions on how to clean it up.

PR #42, opened 2017-05-11
No alias when using GROUPBY on a Function

aweller:
This PR fixes a bug where using GROUP BY on a Function creates an invalid Vertica query if the Function has an alias set.

### DIFF
--- a/pypika/queries.py
+++ b/pypika/queries.py
@@ -542,7 +542,7 @@ class QueryBuilder(Selectable, Term):
     def _where_sql(self, quote_char=None, **kwargs):
         return ' WHERE {where}'.format(where=self._wheres.get_sql(quote_char=quote_char, subquery=True, **kwargs))
 
-    def _group_sql(self, quote_char=None, **kwargs):
+    def _group_sql(self, quote_char=None, with_alias=None, **kwargs):
         return ' GROUP BY {groupby}'.format(
             groupby=','.join(term.get_sql(quote_char=quote_char, **kwargs)
                              for term in self._groupbys)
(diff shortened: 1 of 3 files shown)


pypika/tests/test_groupby_modifiers.py
```diff
@@ -142,3 +142,14 @@ def test_verticaoracle_multiple_rollups(self):
         )
 
         self.assertEqual('SELECT "foo","fiz",SUM("bar") FROM "abc" GROUP BY ROLLUP("foo"),ROLLUP("fiz")', str(q))
+
+
+    def test_verticaoracle_groupby_with_alias(self):
```
twheys: This doesn't belong in the RollupTests class. Can you please move it to `GroupByTests` in `test_selects`?

pypika/queries.py
```diff
@@ -544,7 +544,7 @@ def _where_sql(self, quote_char=None, **kwargs):
 
     def _group_sql(self, quote_char=None, **kwargs):
```
twheys: You should add `with_alias` to the function signature as a kwarg and ignore the value since if it were to be passed in the kwargs it would cause an exception to be raised when you call `term.get_sql`

aweller: Fixed.

pypika/tests/test_groupby_modifiers.py
aweller: Ok, I moved it.

aweller:
All issues should be fixed now.

pypika/tests/test_selects.py
```diff
@@ -210,6 +210,15 @@ def test_groupby__str(self):
 
         self.assertEqual('SELECT "foo",COUNT(DISTINCT *) FROM "abc" GROUP BY "foo"', str(q))
 
+    def test_groupby__select(self):
```
twheys: Could you please change the name of this to indicate what is being tested here?

PR #48, opened 2017-06-08
Added quote_char to Interval class

mikeengland:


pypika/terms.py
```diff
@@ -695,9 +695,10 @@ def get_sql(self, **kwargs):
                 largest=self.largest,
                 smallest=self.smallest,
             ) if self.largest != self.smallest else self.largest
-        return 'INTERVAL \'{expr} {unit}\''.format(
+        return 'INTERVAL {quote}{expr} {unit}{quote}'.format(
```
twheys: quote character is for wrapping names. The single quote is different and shouldn't use quote_char for it.

PR #50, opened 2017-06-08
Added dialects for intervals and specific DB Query classes

mikeengland:


pypika/terms.py
```diff
@@ -701,10 +712,16 @@ def get_sql(self, **kwargs):
                 largest=self.largest,
                 smallest=self.smallest,
             ) if self.largest != self.smallest else self.largest
-        return 'INTERVAL \'{expr} {unit}\''.format(
-            expr=expr,
-            unit=unit,
-        )
+
+        if dialect == Dialects.MYSQL:
```
twheys: instead of using an if branch for this whole part, you could use a dict to map the interval to a string template and then format it.

twheys:
```
interval_templates = {
  Dialects.MYSQL: 'INTERVAL {expr} {unit}',
  Dialects. ORACLE: 'INTERVAL \'{expr}\' {unit}
}

interval_templates.get(dialect, 'INTERVAL \'{expr} {unit}\'')
```

mikeengland: That's a cleaner solution. Will fix 👍

PR #62, opened 2017-07-04
Added negate criterion

aweller:
Added a `NegateCriterion` so each criterion `foo=bar` can be inverted to `NOT foo=bar` by calling `criterion.as_negated()`.

### DIFF
--- a/pypika/terms.py
+++ b/pypika/terms.py
@@ -1,21 +1,61 @@
 # coding: utf8
+import copy
 import re
 from datetime import date
+from functools import partial
 
 from aenum import Enum
 
-from pypika.enums import (Boolean,
-                          Dialects,
-                          Equality,
-                          Arithmetic,
-                          Matching)
-from pypika.utils import (CaseException,
-                          builder,
-                          resolve_is_aggregate,
-                          alias_sql)
+from pypika.enums import (
+    Boolean,
+    Dialects,
+    Equality,
+    Arithmetic,
+    Matching,
+)
+from pypika.utils import (
+    CaseException,
+    builder,
+    resolve_is_aggregate,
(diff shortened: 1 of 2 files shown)


pypika/terms.py
```diff
@@ -101,6 +105,9 @@ def isin(self, arg):
     def notin(self, arg):
         return self.isin(arg).negate()
 
+    def as_negated(self):
```
twheys: can you call this `negate` please

pypika/terms.py
```diff
@@ -351,6 +358,24 @@ def get_sql(self, with_alias=False, **kwargs):
         return sql
 
 
+class NegateCriterion(Criterion):
```
twheys: Please name this class `Not` so the syntax is more natural

pypika/terms.py
twheys: Please remove the negation logic from the `Criterion` class, move it up into term

pypika/terms.py
```diff
@@ -351,6 +358,24 @@ def get_sql(self, with_alias=False, **kwargs):
...
+    def __init__(self, term, alias=None):
```
twheys: please remove alias, term already has it

pypika/terms.py
```diff
@@ -1,23 +1,58 @@
...
+    def fields(self):
+        return self.term.fields() if self.term.fields else []
+
+    def get_sql(self, **kwargs):
+        return "NOT {term}".format(term=self.term.get_sql(**kwargs))
+
+    def __str__(self):
+        return self.get_sql(quote_char='"')
+
+    def __getattr__(self, item):
```
twheys: please add a docstring

PR #69, opened 2017-08-15
Infinite Recursion when using Not criterion

twheys:
Added a catch for deepycopy function calls to the __getattr__ function in the Not class to fix an infinite recursion issue.

### DIFF
--- a/pypika/utils.py
+++ b/pypika/utils.py
@@ -29,13 +29,45 @@ class RollupException(Exception):
 
 
 def builder(func):
+    """
+    Decorator for wrapper "builder" functions.  These are functions on the Query class or other classes used for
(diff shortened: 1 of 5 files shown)


pypika/utils.py
```diff
@@ -57,3 +57,11 @@ def resolve_is_aggregate(values):
 
 def alias_sql(sql, alias, quote_char=None):
     return '{sql} {quote}{alias}{quote}'.format(sql=sql, alias=alias, quote=quote_char or '')
+
+
+def raise_attribute_for_copy(classname, attrname):
+    """
+    This prevents Fields being when deepcopy functions are called by raising an AttributeError
```
aweller: Please fix this docstring.

PR #74, opened 2017-09-27
Added Amazon Redshift dialect and timestamp/date add functions

mikeengland:


pypika/terms.py
```diff
@@ -847,7 +849,7 @@ def fields(self):
         return []
 
     def get_sql(self, **kwargs):
-        dialect = kwargs.get('dialect')
+        dialect = self.dialect or kwargs.get('dialect')
```
twheys: does it make sense to have two sources for dialect? Where do each of them come from? Is it not redundant?

pypika/terms.py
```diff
@@ -880,6 +882,10 @@ def get_sql(self, **kwargs):
             # MySQL requires no single quotes around the expr and unit
             Dialects.MYSQL: 'INTERVAL {expr} {unit}',
 
+            # PostgreSQL and Redshift require quotes around the expr and unit e.g. INTERVAL '1 week'
+            Dialects.POSTGRESQL: 'INTERVAL \'{expr} {unit}\'',
+            Dialects.REDSHIFT: 'INTERVAL \'{expr} {unit}\'',
```
twheys: Could you please add some test coverage for this?

pypika/terms.py
mikeengland: The dialect gets passed to the get_sql() function when the interval is used inside another function. However, just using Interval() by itself doesn't receive the dialect, meaning the default (and incorrect) interval SQL was being returned for the MySQL platform when I was testing.

twheys: ok, sounds like something we should eventually work out, but maybe not in this PR.

PR #81, opened 2017-11-23
Added update and orderby to docs, and a requirements.txt for development

### LINKED ISSUE #76: Add order by example to README
As order by is such a common use case, pypika should include examples of how to add an asc and desc order by.

### LINKED ISSUE #78: Add update example to README
No update example available.

mikeengland:
Locally I had to install aenum and the sphinx packages when developing the code, so I created a requirements.txt to make this more simple.

Fixes #78 Fixes #76

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,5 @@
+aenum==2.0.8
+
+# Documentation
+Sphinx==1.6.5
+sphinx-rtd-theme==0.2.4
--- a/README.rst
+++ b/README.rst
@@ -78,6 +78,18 @@ Both of the above examples result in the following SQL:
 
     SELECT id,fname,lname,phone FROM customers
 
+Results can be ordered by using the following syntax:
+
+.. code-block:: python
+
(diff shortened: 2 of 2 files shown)


requirements.txt
```diff
@@ -0,0 +1,5 @@
+aenum==2.0.8
+
+# Documentation
+Sphinx==1.6.5
```
twheys: can you please make a separate dev requirements file and have setup.py parse these? The dependencies were installed before via setup.py but maybe having requirements.txt is easier.

mikeengland: I'm not sure we need two requirements files (unless I am missing something). The aenum module is required to run the module, and sphinx for the docs (I guess we need them for both dev and prod so that we can build and test them locally).

Maybe we can just have one requirements file and make setup.py parse that?

mikeengland: Nvm - we can keep sphinx in requirements-dev.txt so it doesn't install on setup.py install.

PR #87, opened 2017-12-12
Fixed usage of alias in groupby clause

twheys:
Fixed an issue with aliases in groupby clauses to use the term alias if it is selected in the query

### DIFF
--- a/pypika/tests/test_selects.py
+++ b/pypika/tests/test_selects.py
@@ -2,19 +2,19 @@
 import unittest
 
 from pypika import (
+    Case,
+    Field as F,
+    MSSQLQuery,
+    MySQLQuery,
+    OracleQuery,
+    Order,
+    PostgreSQLQuery,
     Query,
+    RedshiftQuery,
     Table,
     Tables,
-    Field as F,
-    Case,
-    functions as fn,
-    Order,
-    MySQLQuery,
     VerticaQuery,
-    MSSQLQuery,
-    PostgreSQLQuery,
-    OracleQuery,
-    RedshiftQuery,
+    functions as fn,
 )
 
 __author__ = "Timothy Heys"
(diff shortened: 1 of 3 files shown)


pypika/tests/test_selects.py
```diff
@@ -263,15 +264,26 @@ def test_groupby__str(self):
...
-
-        self.assertEqual('SELECT SUM("foo") "bar" FROM "abc" GROUP BY SUM("foo")', str(q))
-
-    def test_groupby_with_case_does_not_show_an_alias(self):
+        bar = self.t.bar.as_('bar01')
+        q = Query.from_(self.t) \
+            .select(fn.Sum(self.t.foo), bar) \
+            .groupby(bar)
+
+        self.assertEqual('SELECT SUM("foo"),"bar" "bar01" FROM "abc" GROUP BY "bar01"', str(q))
```
inkmix: is there also a test where a column has an alias but the query is grouped by the column name (not the alias)? if not, might be worth adding, I think.

twheys: yes, most of the group by tests

inkmix: hmm, I now looked through all of them with aliases in selects and we always use the alias in the group by clause as well.
what I meant is having an alias in a select and then using the original column name in the group by, e.g. (I think, I'm still not very familiar with the syntax):
```
bar = self.t.bar.as_('bar01')
          q = Query.from_(self.t) \
              .select(fn.Sum(self.t.foo), bar) \
              .groupby(self.t.bar)
```

twheys: In your example it will group by `SUM("foo")` because it uses the alias to determine if the alias can be used in the group by.

inkmix: right!, probably without the SUM in the select is also fine. either way, I think it's a good case to add

PR #93, opened 2017-12-27
Disable union wrap for clickhouse

nibrag:


twheys:
bumped version to 0.10.1

PR #94, opened 2017-12-29
Fixed null wrapping in update set

Zeliboba5:


twheys:
bumped version to 0.10.2

PR #99, opened 2018-01-29
Changed rollup to combine additional rollup fields

twheys:
Changed the behavior of rollup to combine subsequent rollup fields after a rollup has been added to the group by clause

### DIFF
--- a/pypika/queries.py
+++ b/pypika/queries.py
@@ -359,12 +359,17 @@ class QueryBuilder(Selectable, Term):
                  for term in terms]
 
         if for_mysql:
+            # MySQL rolls up all of the dimensions always
             if not terms and not self._groupbys:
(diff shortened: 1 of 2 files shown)


pypika/queries.py
```diff
@@ -366,6 +366,8 @@ def rollup(self, *terms, **kwargs):
             self._mysql_rollup = True
             self._groupbys += terms
 
+        elif 0 < len(self._groupbys) and isinstance(self._groupbys[-1], Rollup):
```
mikeengland: I think a comment here would be handy to know the difference between this branch and the one below it.

twheys: fixed

mikeengland: Thanks

PR #100, opened 2018-03-06
Bugfixing

twheys:
- Fixed a bug where joined queries would have their alias overwritten
- Fixed ordering in union queries
- Added alias support to ValueWrapper
- Fixed the subquery in FROM clause to not overwrite the alias
- Fixed union queries to work with multiple queries
- Fixed UnionQuery to support limit and offset

### DIFF
--- a/pypika/queries.py
+++ b/pypika/queries.py
@@ -307,7 +307,7 @@ class QueryBuilder(Selectable, Term):
 
         self._from.append(Table(selectable) if isinstance(selectable, str) else selectable)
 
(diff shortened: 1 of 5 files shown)


pypika/queries.py
```diff
@@ -170,6 +172,107 @@ def update(cls, table):
         return cls._builder().update(table)
 
 
+class UnionQuery(Selectable, Term):
+    def __init__(self, base_query, union_query, union_type, alias=None):
```
mikeengland: This could do with a docstring

twheys: Will add

PR #102, opened 2018-03-08
Added not like function for negating wildcard queries

mikeengland:
Added not like function for negating wildcard queries

### DIFF
--- a/pypika/tests/test_selects.py
+++ b/pypika/tests/test_selects.py
@@ -208,6 +208,21 @@ class WhereTests(unittest.TestCase):
 
         self.assertEqual("SELECT * FROM \"abc\" WHERE \"foo\" LIKE 'ab%'", str(q))
 
+    def test_where_field_contains(self):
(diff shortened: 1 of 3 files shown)


pypika/tests/test_selects.py
```diff
@@ -208,25 +208,25 @@ def test_where_field_starts_with(self):
 
         self.assertEqual("SELECT * FROM \"abc\" WHERE \"foo\" LIKE 'ab%'", str(q))
 
-    def test_where_field_contains(self):
-        q = Query.from_(self.t).select(self.t.star).where(self.t.foo.like('%fg%'))
+    def test_where_field_does_not_start_with(self):
```
x8lucas8x: Wait, so you replaced all the like tests with just not like ones? Shouldn't we keep both?

mikeengland: Thanks for catching. That was not intentional.

PR #103, opened 2018-03-16
RETURNING clause for postgresql

Zeliboba5:
This PR implements postgres `RETURNING` clause.
This PR implements it directly in QueryBuilder, which could be quite misleading, for other sql users, so if you have any propositions regarding how this code should be arranged I can do some refactoring

### DIFF
--- a/pypika/queries.py
+++ b/pypika/queries.py
@@ -301,7 +301,6 @@ class QueryBuilder(Selectable, Term):
         self._wheres = None
         self._prewheres = None
         self._groupbys = []
-        self._returns = []
         self._with_totals = False
         self._havings = None
         self._orderbys = []
@@ -314,7 +313,6 @@ class QueryBuilder(Selectable, Term):
         self._updates = []
 
         self._select_star = False
-        self._return_star = False
         self._select_star_tables = set()
         self._mysql_rollup = False
         self._select_into = False
@@ -445,18 +443,6 @@ class QueryBuilder(Selectable, Term):
             self._havings = criterion
 
     @builder
-    def returning(self, *terms):
-        for term in terms:
-            if isinstance(term, Field):
-                self._return_field(term)
-            elif isinstance(term, str):
-                self._return_field_str(term)
-            elif isinstance(term, (Function, ArithmeticExpression)):
-                self._return_other(term)
-            else:
-                self._return_other(self._wrap(term))
-
-    @builder
     def groupby(self, *terms):
         for term in terms:
             if isinstance(term, str):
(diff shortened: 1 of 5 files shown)


twheys:
If you'd like to add a vendor specific function you can override the querybuilder class and instantiate it in the PostgreSQLQuery class' _builder function. See my last PR for an example: https://github.com/kayak/pypika/pull/109

pypika/queries.py
```diff
@@ -570,6 +584,49 @@ def _select_other(self, function):
...
+        elif self._update_table:
+            self._return_field(Field(term, table=self._update_table))
+        elif self._delete_from:
+            self._return_field(Field(term, table=self._from[0]))
+        else:
+            raise QueryException('Returning can\'t be used in this query')
+
+    def _validate_returning_term(self, term):
+        if not any([self._insert_table, self._update_table, self._delete_from]):
+            QueryException('Returning can\'t be used in this query')
```
twheys: It looks like you're missing the `raise` keyword to raise this exception

pypika/queries.py
```diff
@@ -570,6 +584,49 @@ def _select_other(self, function):
...
+        if (
+                term.table not in {self._insert_table, self._update_table}
+                and term not in self._from
+        ):
+            QueryException('You can\'t return from other tables')
```
twheys: It looks like you're missing the `raise` keyword to raise this exception

pypika/queries.py
```diff
@@ -442,6 +444,18 @@ def having(self, criterion):
         else:
             self._havings = criterion
 
+    @builder
+    def returning(self, *terms):
```
twheys: This code looks very similar to the `select` function. Would it be possible to consolidate the similarities between the two?

pypika/tests/test_inserts.py
```diff
@@ -88,6 +88,35 @@ def test_insert_null(self):
         self.assertEqual('INSERT INTO "abc" VALUES (NULL)', str(query))
 
 
+class InsertIntoReturningTests(unittest.TestCase):
```
twheys: These test cases look pretty thorough and I'm not sure what's missed but coveralls is reporting 0.5% drop in test case coverage. Can you please check and make sure all of your conditions are covered by test cases?

twheys:
Sorry, not sure how I closed this, didn't mean to. Re-opened.

Zeliboba5:
Moved everything to postgres QueryBuilder, added some more tests and covered some more clauses, now it all looks neat.
Regarding consolidation of `returning` and `select` I think it is wrong way to go, as it still different functions, which are not fully compatible (`RETURNING` does not support aggregate functions, so I covered it in my new commits)

twheys:
Looks good to go, just need to fix the merge conflicts. I'll merge once that's done, let me know if you need help with it.

twheys:
Merged and published 0.12.0

PR #110, opened 2018-03-26
Fixed single quote escaping

Zeliboba5:
I looked through supported dialects and it looks like all of them escape single quote by doubling it, so it should be okay without any dialect passing to .get_sql() shenanigans

### DIFF
--- a/pypika/terms.py
+++ b/pypika/terms.py
@@ -222,7 +222,8 @@ class ValueWrapper(Term):
         if isinstance(self.value, date):
             return "'%s'" % self.value.isoformat()
         if isinstance(self.value, basestring):
-            return "'%s'" % self.value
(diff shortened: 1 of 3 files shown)


pypika/terms.py
```diff
@@ -222,7 +222,8 @@ def _get_value_sql(self):
         if isinstance(self.value, date):
             return "'%s'" % self.value.isoformat()
         if isinstance(self.value, basestring):
-            return "'%s'" % self.value
+            value = self.value.replace("'", "''")
+            return "'%s'" % value
```
twheys: Ah, this should code eventually escape all SQL characters, but haven't managed to find time to implement that yet.

twheys:
Merged and published version 0.12.1

PR #116, opened 2018-04-03
implement ON DUPLICATE KEY UPDATE for MySQL

inkmix:
* I've chosen to have a similar syntax to `.set` for `.on_duplicate_key_update` to keep updating values consistent
* there's a new `Values` wrapper to represent the `VALUES` term in the new clause

### DIFF
--- a/pypika/terms.py
+++ b/pypika/terms.py
@@ -210,13 +210,15 @@ class ValueWrapper(Term):
         return []
 
     def get_sql(self, quote_char=None, **kwargs):
-        sql = self._get_value_sql()
+        sql = self._get_value_sql(quote_char=quote_char, **kwargs)
         if self.alias is None:
             return sql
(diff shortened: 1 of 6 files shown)


pypika/terms.py
```diff
@@ -231,6 +233,18 @@ def _get_value_sql(self):
...
+class Values(Term):
+    def __init__(self, field,):
+        super(Values, self).__init__(None)
+        self.field = Field(field) if not isinstance(field, Field) else field
+
+    def fields(self):
+        return []
+
+    def get_sql(self, quote_char=None, **kwargs):
+        return f'VALUES({self.field.get_sql(quote_char=quote_char, **kwargs)})'
```
twheys: Can't use a format string here, this needs to support python 2.7 and 3+

inkmix: true thanks. and already noticed thanks to the build 👍

PR #118, opened 2018-04-16
Changed `Not` to inherit from Criterion

Zeliboba5:
Currently `Not` does not inherit from `Criterion`, and because of it you can't build complex criterion from negated term, even though it is valid in sql


twheys:
Thanks for the PR. Published 0.14.1.

PR #131, opened 2018-06-14
Add for_ to base term class.

tobymao:
for_ is not defined for all terms, like ValueWrapper. This adds a noop
to the base for_ definition for convenience.

When running for_ on a criterion, Field("x") > 1, this crashes due to
ValueWrapper not having it implemented.

### DIFF
--- a/pypika/terms.py
+++ b/pypika/terms.py
@@ -69,6 +69,17 @@ class Term(object):
 
         return ValueWrapper(val)
 
+    def for_(self, table):
+        """
+        Replaces the tables of this term for the table parameter provided.  The base implementation returns self because not all terms have a table property.
+
+        :param table:
+            The table to replace with.
+        :return:
+            Self.
+        """
+        return self
+
     def fields(self):
         return [self]
 
@@ -678,9 +689,7 @@ class Function(Term):
         :return:
             A copy of the field with it's table value replaced.
         """
-        self.args = [param.for_(table) if hasattr(param, 'for_')
-                     else param
-                     for param in self.args]
+        self.args = [param.for_(table) for param in self.args]
 
     def get_special_params_sql(self, **kwargs):
         pass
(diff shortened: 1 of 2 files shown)


pypika/terms.py
```diff
@@ -678,9 +689,7 @@ def for_(self, table):
         :return:
             A copy of the field with it's table value replaced.
         """
-        self.args = [param.for_(table) if hasattr(param, 'for_')
```
twheys: Wouldn't this condition here result in just using the instance of ValueWrapper since it doesn't implement `for_` ? Looks to me like this condition intended to implement this behavior already. What about it was broken?

tobymao: criterions were broken because for_ is not implemented. without this fix, my test would fail.

this condition is only on Function. I simply cleaned it up because it is not needed anymore. I felt like it was cleaner to implement on the base class than to have to implement this check on every Term object.

twheys: I mean the assumption is that if `for_` is not implemented then you don't need to replace the table for any contained terms

twheys: I think the issue is in the implementation of `BasicCriterion.for_`

```
    @builder
    def for_(self, table):
        self.left = self.left.for_(table)
        self.right = self.right.for_(table)
```

Looks like here it does not check if for is implemented as it does in `Term`, although i see your point, that you would have to check everywhere if it is implemented or not.

twheys: Going to approve and merge your changes

tobymao: What if you have something like Criterion(self.table.a > 1) 

you want to do Criterion.for_(new_table). this will crash because the value wrapper doesn't implement for_

tobymao: sorry if i wasn't clear. thanks!

twheys:
Published 0.14.5 including these changes

PR #161, opened 2018-09-26
Replaced the usage of deepcopy

### LINKED ISSUE #160: copy.deepcopy() takes 55% of runtime of a simple benchmark
Hi, we are working on [Tortoise ORM](https://github.com/tortoise/tortoise-orm) and I did a series of simple benchmarks to start finding performance bottlenecks. In the one transactioned insert test:
https://github.com/tortoise/orm-benchmarks#interesting-profiling-results
I found that `copy.deepcopy()` takes ~55% of the runtime. Changing it to `copy.copy()` gives me an 83% speed up in that particular test.

Of course it is there for a reason as without it we can get unexpected side-effects, such as what the `pypika.tests.test_joins.JoinBehaviorTests.test_immutable__queries_after_join` test determines.

I'm just wondering what can we do to speed this up?

twheys:
Replaced the usage of deepcopy in the @builder decorator with copy and overrode __copy__ in QueryBuilder

Fixes #160

### DIFF
--- a/pypika/queries.py
+++ b/pypika/queries.py
@@ -9,7 +9,7 @@ from pypika.utils import (
     UnionException,
     alias_sql,
     builder,
-    ignoredeepcopy,
+    ignore_copy,
 )
 from .terms import (
     ArithmeticExpression,
@@ -37,7 +37,7 @@ class Selectable(object):
     def star(self):
         return Star(self)
 
-    @ignoredeepcopy
+    @ignore_copy
     def __getattr__(self, name):
         return self.field(name)
 
@@ -332,6 +332,14 @@ class QueryBuilder(Selectable, Term):
         self.quote_char = quote_char
         self.dialect = dialect
         self.wrap_union_queries = wrap_union_queries
+
+    def __copy__(self):
+        newone = type(self)()
+        newone.__dict__.update(self.__dict__)
+        for key, value in self.__dict__.items():
+            if isinstance(value, (set, list)):
+                newone.__dict__[key] = type(value)(x for x in value)
+        return newone
 
     @builder
     def from_(self, selectable):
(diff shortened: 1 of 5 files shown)


pypika/queries.py
```diff
@@ -333,6 +333,15 @@ def __init__(self, quote_char='"', dialect=None, wrap_union_queries=True):
         self.dialect = dialect
         self.wrap_union_queries = wrap_union_queries
 
+    def __copy__(self):
+        newone = type(self)()
+        newone.__dict__.update(self.__dict__)
+        for attr in self.__dict__.keys():
+            value = self.__dict__[attr]
+            if isinstance(value, list):
```
mikeengland: Is there also a case when the value can be another collection e.g. a tuple, or is it always a list?

pypika/queries.py
```diff
@@ -333,6 +333,15 @@ def __init__(self, quote_char='"', dialect=None, wrap_union_queries=True):
         self.dialect = dialect
         self.wrap_union_queries = wrap_union_queries
 
+    def __copy__(self):
+        newone = type(self)()
+        newone.__dict__.update(self.__dict__)
+        for attr in self.__dict__.keys():
```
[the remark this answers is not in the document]
twheys: no, just wasn't paying attention... will fix it

pypika/queries.py
twheys: well in the QueryBuilder class there is one attribute that is a set, so I'll add that here. Tuples are immutable, so it wouldn't be a problem to share them across copies of a query

mikeengland: Ok, as long as that collection always stores one level of immutable objects that is fine. I didn't know if it may store something mutable e.g. a nested list which you'd need to deep copy.

PR #184, opened 2018-11-14
Schema class

twheys:
Added a Schema class and support for it on Functions. Also added usage of it to the Table class.

### DIFF
--- a/pypika/queries.py
+++ b/pypika/queries.py
@@ -1,4 +1,6 @@
 # coding: utf-8
+from functools import reduce
+
 from pypika.enums import (
     JoinType,
     UnionType,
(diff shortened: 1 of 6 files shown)


pypika/queries.py
```diff
@@ -66,29 +69,51 @@ def __hash__(self):
...
-                             for part in parts)
+class Table(Selectable):
+    def __init__(self, name, schema=None, alias=None):
+        super(Table, self).__init__(alias)
+        self._table_name = name
+
+        # This is a bit complicated in order to support backwards compatibility. It should probably be cleaned up for
+        # the next major release. Schema is accepted as a string, list/tuple, Schema instance, or None
+        self._schema = schema \
+            if isinstance(schema, Schema) \
```
mikeengland: I normally prefer anything more than a single conditional expression to be split out into multiple if statements on different lines for readability

twheys: pulled it out into it's own function

PR #185, opened 2018-11-14
Fixed Query.select('field') with no from table.

### LINKED ISSUE #140: Descriptive error when .from_() call is omitted
I just spent some time figuring out what happened when I transcribed a slightly comples query from raw SQL to pypika (this is a great library, BTW!). The error I was getting was quite cryptic:

```
  File "<stdin>", line 1, in <module>
  File "/home/butla/.virtualenvs/bla-Y9MmHOr2/lib/python3.6/site-packages/pypika/queries.py", line 161, in select
    return cls._builder().select(*terms)
  File "/home/butla/.virtualenvs/bla-Y9MmHOr2/lib/python3.6/site-packages/pypika/utils.py", line 46, in _copy
    result = func(self_copy, *args, **kwargs)
  File "/home/butla/.virtualenvs/bla-Y9MmHOr2/lib/python3.6/site-packages/pypika/queries.py", line 379, in select
    self._select_field_str(term)
  File "/home/butla/.virtualenvs/bla-Y9MmHOr2/lib/python3.6/site-packages/pypika/queries.py", line 565, in _select_field_str
    self._select_field(Field(term, table=self._from[0]))
IndexError: list index out of range
list index out of range
```

Turns out I forgot to add `from_()` call.

**Steps to reproduce:**
run `Query.select('asdf')`

twheys:
Fixes #140 Added a check when selecting from a field using a string argument for the field name when there are no tables currently being selected from in the query

### DIFF
--- a/pypika/tests/test_selects.py
+++ b/pypika/tests/test_selects.py
@@ -11,6 +11,7 @@ from pypika import (
     Order,
     PostgreSQLQuery,
     Query,
+    QueryException,
     RedshiftQuery,
     Table,
     Tables,
@@ -29,6 +30,15 @@ class SelectTests(unittest.TestCase):
         q = Query.from_('abc')
 
         self.assertEqual('', str(q))
+
+    def test_select_no_from(self):
+        q = Query.select(1)
+
+        self.assertEqual('SELECT 1', str(q))
+
+    def test_select_no_from_with_field_raises_exception(self):
+        with self.assertRaises(QueryException):
+            Query.select('asdf')
 
     def test_select__star(self):
         q = Query.from_('abc').select('*')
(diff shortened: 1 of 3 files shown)


pypika/tests/test_selects.py
```diff
@@ -30,6 +31,15 @@ def test_empty_query(self):
 
         self.assertEqual('', str(q))
 
+    def test_select_no_from(self):
+        q = Query.select(1)
+
+        self.assertEqual('SELECT 1', str(q))
+
+    def test_select_no_from_with_field_raises_exception(self):
```
mikeengland: Just wondering - will this stop queries such as `select @@version` in MySQL?

twheys: Don't think that would've been possible to begin with.

PR #189, opened 2018-11-19
Fix table sql with schema and alias not showing alias.

tobymao:


twheys:
Published in 0.18.3. Thanks for spotting and fixing!

tobymao:
@twheys my project that is using pypika is still failing, i'll investigate, but i think there are some other issues.

tobymao:
turns out it was because i was passing empty string as schema, that was translating into ""."table" -- i think this is probably fine behavior, i've changed my code to use None instead.

PR #193, opened 2018-12-12
[RFC] Added option to not use aliases in groupby and orderby

diegobatt:
The GROUP BY and ORDER BY clauses were using the aliases defined in the SELECT statement, this is not ANSI SQL and hence it does not work with every SQL implementation. Now in order to not parse the query with aliases in the `group by` and `order by` one could use `get_sql(groupby_alias=False, orderby_alias=False)`. Since not using this aliases is ANSI, I think that False should be the default behavior, but I'm keeping it as True in order to don't break things.


diegobatt:
@twheys I added some tests to both the orderby and groupby.

twheys:
great, thank you. Looks good. Will merge and publish.

PR #197, opened 2019-01-04
Bracket term

twheys:
Added a term for holding brackets in order to maintain order of operations in expressions


gautham20:
kindly add `Term.Bracket` to the docs.

PR #198, opened 2019-01-04
Empty Criterion

### LINKED ISSUE #196: Add EmptyCriterion for easier generic queries
There are cases when you are building query complex criterion based on some external checks.
And in those cases you could want to have some empty criterion to build you complex criterion from it.

Basic case, where you have dict of argument and want to have any corresponding rows, could look like this:
```
criterion = EmptyCriterion()

for key, value in filters.items():
    criterion |= (table.key == value)

query = query.where(criterion)
```

Which right now would look like this:
```
criterion = None

for key, value in filters.items():
    if not criterion:
        criterion = (table.key == value)
    else:
        criterion |= (table.key == value)

if criterion:
    query = query.where(criterion) 
```

Also, there could be more complex cases where you got two criterions from some generic functions, and there you would have to do:
```
if not criterion1 and criterion2:
    res_criterion = criterion2
elif criterion1 and not criterion2:
    res_criterion = criterion1
else:
   res_criterion = criterion1 | criterion2
```
While, if functions could return EmptyCriterion, you could just do:
```
res_criterion = criterion1 | criterion2
```

Does this seem like good idea for you? If you approve this design I could try implementing it

twheys:
Fixes #196 Added an EmptyCriterion class which allows the user to build up boolean clauses using lists of constituents. Added any and all static methods to Criterion to give an easy syntax for using this.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -257,6 +257,33 @@ XOR
     SELECT id,fname,lname,phone FROM customers WHERE age>=18 XOR is_registered
 
 
+Convenience Methods
+"""""""""""""""""""
+
+In the `Criterion` class, there are static methods that allow building chains AND and OR expressions with a list of
+terms.
+
+.. code-block:: python
+
+    from pypika import Criterion
+
+    customers = Table('customers')
+    q = Query.from_(customers).select(
+        customers.id,
+        customers.fname
+    ).where(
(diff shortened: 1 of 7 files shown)


abondar:
Seems all good to me 👍

PR #213, opened 2019-02-05
Rename 'Psueudocolumns' class and fix functionality

### LINKED ISSUE #212: How to include SQL_CALC_FOUND_ROWS option?
I have a query that I'm trying to replicate in pypika that uses the `SQL_CALC_FOUND_ROWS` [1] option, but cannot figure out a way to get the option in the query. If I add it as a column, pypika wants to select it, but the option shouldn't be quoted in any way and there shouldn't be a comma to separate the option from the columns.

```python
purchase_order = pypika.Table("purchase_order")
query = (
        pypika.MySQLQuery.from_(purchase_order)
        .where(purchase_order.is_active == 1)
        .select("SQL_CALC_FOUND_ROWS")
        .select(purchase_order.purchase_order_id.as_("_id"))
        .limit(pypika.Parameter("%(limit)s"))
    )
```

will result in 

```sql
SELECT 
  `SQL_CALC_FOUND_ROWS`,
  `purchase_order_id` `_id`
FROM `purchase_order` 
WHERE `is_active`=1 
LIMIT %(limit)s
```

however the correct syntax for this option is...

```sql
SELECT 
  SQL_CALC_FOUND_ROWS
  `purchase_order_id` `_id`
FROM `purchase_order` 
WHERE `is_active`=1 
LIMIT %(limit)s
```

Is there a way I can insert this option? I saw a `Psuedocolumn` class that _seems_ like the right area, but if I try defining a new `Psuedocolumn` and passing it in, I get an exception raised.

```python
purchase_order = pypika.Table("purchase_order")
CalcFoundRows = pypika.terms.Psuedocolumn('SQL_CALC_FOUND_ROWS')
query = (
        pypika.MySQLQuery.from_(purchase_order)
        .where(purchase_order.is_active == 1)
        .select(CalcFoundRows)
        .select(purchase_order.purchase_order_id.as_("_id"))
        .limit(pypika.Parameter("%(limit)s"))
    )
```

raises...
```
AttributeError
'Psuedocolumn' object has no attribute 'table'
```

Any pointers on how I could get this to work? This is a fantastic lib by the way, I needed a way to programmatically define queries and *really* didn't want to throw an ORM into the mix. Thanks.

[1] https://dev.mysql.com/doc/refman/5.7/en/information-functions.html#function_found-rows

mikeengland:
Rename 'Psueudocolumns' class to use correct spelling and ensure it can be used in queries without quotes

Fixes #212 

@stickystyle does this now offer the functionality you require?


pypika/tests/test_pseudocolumns.py
```diff
@@ -0,0 +1,57 @@
...
+    SysDate,
+)
+from pypika.terms import PseudoColumn
+
+
+class PseudocolumnsTest(unittest.TestCase):
+
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
```
x8lucas8x: This will break for v2 build.

pypika/tests/test_pseudocolumns.py
```diff
@@ -0,0 +1,57 @@
...
+    ObjectID,
+    ObjectValue,
+    RowID,
+    RowNum,
+    SysDate,
+)
+from pypika.terms import PseudoColumn
+
+
+class PseudocolumnsTest(unittest.TestCase):
```
x8lucas8x: Minor: This is not camel cased.

PR #216, opened 2019-02-06
Add query modifiers functionality to MySQLQueryBuilder

mikeengland:
This allows query modifiers like HIGH_PRIORITY and SQL_CALC_FOUND_ROWS to be added to MySQL queries.


mikeengland:
Released as v0.24.0

PR #222, opened 2019-03-08
Add support for APPROX_COUNT_DISTINCT

philip-newzoo:
Some SQL engines have a HyperLogLog implementation of count distinct.

This PR adds support for this function. 

See [this article](https://www.red-gate.com/simple-talk/sql/t-sql-programming/there-is-a-new-count-in-town/) for more details.


twheys:
Hi, thanks for the PR. The functions provided in PyPika are only the most commonly used SQL functions. If we tried to include every single SQL function, it would be a nightmare to manage. Please implement this function in your own project.

PR #225, opened 2019-03-12
Update queries.py

vorlon001:
add new function order_as, limitr

example:
import pypika
customers = pypika.Table('customers')
q = pypika.MySQLQuery.from_(customers).select(
customers.id, customers.fname, customers.lname, customers.phone, "sdfas"
).where(
((customers.age >= 18) | (customers.age < 10)) & (customers.lname == 'Mustermann')
).groupby(customers.x).order_as('DESC').orderby(customers.y).limitr(11,22)
print(str(q))

output:
SELECT id,fname,lname,phone,sdfas FROM customers WHERE (age>=18 OR age<10) AND lname='Mustermann' GROUP BY x ORDER BY y as DESC LIMIT 11, 22

### DIFF
--- a/pypika/queries.py
+++ b/pypika/queries.py
@@ -246,7 +246,13 @@ class _UnionQuery(Selectable, Term):
         self._unions = [(union_type, union_query)]
         self._orderbys = []
 
+        self._ordertype = None
+
         self._limit = None
+        
+        self._limit_min = None
+        self._limit_max = None
+
         self._offset = None
 
         self._wrapper_cls = wrapper_cls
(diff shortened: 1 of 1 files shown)


pypika/queries.py
```diff
@@ -374,11 +397,18 @@ def __init__(self, quote_char='"', dialect=None, wrap_union_queries=True, wrappe
...
+
         self._orderbys = []
+        self._ordertype = None
+
         self._joins = []
         self._unions = []
 
         self._limit = None
+
+        self._limit_min = None
```
twheys: instead of adding these as extra fields, I think  it would be reasonable to recycle the existing limit function and pass a tuple as the argument to it. Wdyt?

vorlon001: yes.
I added this feature so as not to mess up the code.

Issue #227, opened 2019-03-19
Allow filtering by non existant table


jtrain:
When creating a subquery, often I'd like to reference the parent table id.
However this validation code https://github.com/kayak/pypika/blob/master/pypika/queries.py#L718 prevents that.

For a contrived example

```
Teacher, School = Tables("teacher", "school")
Query.from_(School).select(
 Query.from_(Teacher).select(Teacher.name).where(
  Teacher.current_school_id==School.id
  ).limit(1).as_('a_teacher')
)
```

Here Teacher is referencing the parent 'School' table and the validation would fail.

My use case is generally including a JSONB array of tags or similar from the result of a subquery in the parent.

I have worked around this by adding the table, filtering the subquery, then removing the table. Kind of like a decorate / action / undecorate pattern.

Options I can think of for solving this:
1. A flag that would disable this particular validation check (but still allow others);
2. An API for a validations framework where validations can be added and removed;
3. A way of telling a subquery about a known table, that would pass this validation and have no effect on output SQL otherwise; and
4. Continue the decorate / action /undecorate pattern in our code.

I'm happy to develop a PR if any of these, or another option excites you.

twheys:
Appreciate the suggestions. Might be beneficial to have configurable validation, but in most cases pypika avoids any sort of validation, leaving that to the database itself. In this case, it might be that pypika needs to reference something in the table, so it enforces that the table is actually selected from. This is a case that isn't considered, but the subquery does know it's a subquery when the SQL is being generated.

Maybe as a short term solution we could take the first option and make the `where` function accept a kwarg to disable the validation?

jtrain:
Sure I'll take a look, and see what feels good.

jtrain:
#228 An approach for fixing this, other approaches could easily be taken if that wasn't the best trade off

PR #228, opened 2019-03-20
Adds support for correlated subqueries

jtrain:
Correlated subqueries [1] were prevented by validation which asserted
that the field filtered existed in the `_from` tables. There was a work
around which was to temporarily add the table to the _from list, then
create the correlated subquery, then later remove it.

This approach introduces a way to specifically highlight that the query
being built is a correlated subquery, and as such include the outer
table at the time of creation with `from_`.

An alternate approach may have been to introduce a flag which disabled
table based validation on fields when creating new WHERE clauses. This
approach was decided against because a correlated subquery requires
namespaces lest there be a clash when filtering "id" against "id" in the
outer table.

[1] https://en.wikipedia.org/wiki/Correlated_subquery

### DIFF
--- a/pypika/queries.py
+++ b/pypika/queries.py
@@ -168,7 +168,7 @@ class Query(object):
         return QueryBuilder()
 
     @classmethod
-    def from_(cls, table):
+    def from_(cls, table, correlated_table=None):
         """
         Query builder entry point.  Initializes query building and sets the table to select from.  When using this
         function, the query becomes a SELECT query.
@@ -178,9 +178,16 @@ class Query(object):
 
             An instance of a Table object or a string table name.
 
+        :param correlated_table:
+            Type: ``Table``, ``Query``, or ``str``
+
+            An instance of the outer table (or string table name)
+            used if this query is forming part of a correlated subquery.
+            (a subquery inside SELECT or WHERE).
+
         :returns QueryBuilder
         """
-        return cls._builder().from_(table)
+        return cls._builder().from_(table, correlated_table)
 
     @classmethod
     def into(cls, table):
@@ -358,6 +365,7 @@ class QueryBuilder(Selectable, Term):
         super(QueryBuilder, self).__init__(None)
 
         self._from = []
+        self._correlated = []
         self._insert_table = None
         self._update_table = None
         self._delete_from = False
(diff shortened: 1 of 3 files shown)


jtrain:
I am not wedded to the API designed here. There are some downsides:

1. You may want to reference more than one correlated outer table
2. You are required to set a `from_` table because this reuses that same method and the selectable isn't optional

An alternate point of view may be to create a new API for relating to outer tables. For example:

```
Query.from_(history).correlated(customers)
```

This way you could introduce many such correlated tables, either by an args list for `correlated` or chained calls to it.

It would be simple to change to this new API, but i was reluctant to make a big change considering this isn't my project

jtrain:
This is to address #227 and since creating that ticket I learned the name of the SQL feature I was trying to implement (Correlated subquery)

pypika/tests/test_selects.py
```diff
@@ -157,6 +157,22 @@ def test_select__nested_subquery(self):
                          'FROM (SELECT "sq0"."foo","sq0"."bar" '
                          'FROM (SELECT * FROM "abc") "sq0") "sq1") "sq2"', str(q))
 
+    def test_correlated_subquery(self):
+        subquery = Query.from_(
+            self.table_abc, correlated_table=self.table_efg
```
twheys: I'm wondering if we can make this a bit more implicit, as it is in SQL, although that's tricky because python is always going to  interpret the subquery expression first. Perhaps we could just remove validation on the table entirely?

pypika/queries.py
```diff
@@ -423,12 +432,27 @@ def from_(self, selectable):
...
         :returns
             A copy of the query with the table added.
         """
 
         self._from.append(Table(selectable) if isinstance(selectable, str) else selectable)
 
+        if correlated_table is not None:
+            # add a reference to the outer table in a correlated subquery.
+            self._correlated.append(
+                Table(correlated_table)
```
twheys: Need to consider the case that the correlated value is another Query, not a Table. Should probably add a test for that case. Instances of Tables and Queries always need to be interchangeable.

jtrain:
I'm not in love with spelling out the correlated table. Removing the validation would be great, and in line with the stated goal!

Let's say I was writing a query directly in text. I could also have a typo in the table name, or reference a table that doesn't exist. The database will error out and I'll know about it.

Should I close this branch and open a new one for removing the validation? If we needed to keep it around for legacy code, then it could be removed via a flag as you originally suggested. I didn't want to go down that path if there were going to be new validations added in the future, because we'd end up with as many flags.

jtrain:
What are your thoughts on exploring simply removing the validation?

twheys:
yeah, I think that's the route to go. If you give me a couple of days I can open a PR for it, or you're welcome to update yours with that change.

jtrain:
I'll make a new clean one. There is no rush on my end, I'm using the work around dance I posted earlier, just happy to contribute back.

PR #230, opened 2019-03-20
Documents design goals for |Brand| er pypika

jtrain:
A section up top so people are in the right frame of mind.

I tried to emphasis the trade offs that were being made, while we don't hvae validation, we do have fast query building.


twheys:
Appreciate the contribution. This is definitely the stance we take, and I agree that it's important to make it clear in the docs.

PR #237, opened 2019-03-26
Fixes #211 Added support for pseudocolumns and oracle ROWNUM

### LINKED ISSUE #211: Oracle SQL limit syntax is wrong
If I want that only x entries will be returned then I could use `.limit(x)` but for Oracle the syntax isn't correct.

Using `OracleQuery` and `.limit(2)` results in

`'SELECT * FROM "table" LIMIT 2'`

but that doesn't work in Oracle DB.

Correct would be:

`'SELECT * FROM "table" WHERE ROWNUM <= 2'`

twheys:


mikeengland:
By the way, we already have PseudoColumn support (see the bottom of terms.py and pseudocolumns.py). RowNum is already part of this.

twheys:
yeah i noticed this from the other issue that was raised. I will consolidate them.

PR #251, opened 2019-05-20
Fix error in correlated subquery documentation

jtrain:
The Python example didn't appear because it was formatted incorrectly.


mikeengland:
Thanks for fixing!

PR #252, opened 2019-05-20
Fix NULLIF syntax so it generates functional SQL

akmiller01:
The correct syntax for NULLIF is:
```
SELECT NULLIF("foo",0) FROM "abc"
```
Instead of:
```
SELECT NULLIF("foo"=0) FROM "abc"
```


twheys:
Thanks for fixing this. Will publish a new version with the fix soon.

PR #256, opened 2019-06-05
Fix PostgreSQL arrays

xmsid:


pypika/dialects.py
```diff
@@ -273,6 +274,15 @@ def get_sql(self, with_alias=False, subquery=False, **kwargs):
             querystring += self._returning_sql()
         return querystring
 
+    def _values_sql(self, **kwargs):
```
twheys: Instead of introducing a special case here, it would make sense to move it to the Array class. Imagine if we had 5 other special cases like this that we tried to handle here, it would be chaos.

One of the kwargs passed into the `get_sql` function tells you the sql dialect. You can key off that value for postgres and return the correct sql.

twheys: Also, please add a test case to cover this.

xmsid: Should I make a `test_postgresql.py` in the `dialects/` folder? or should I add it to `test_inserts.py`?

twheys: I don't have a strong opinion on that. There might be some examples you can reference for special mysql cases, if you want to make it consistent with those.

xmsid: I added a test case for it in `dialects/test_postgresql.py` but it seems that `setup.py tests` does not run the test cases in the `dialects/` folder?

PR #257, opened 2019-06-08
Support instantiate many tables with alias

PinaGamer:
In order to be faster and easier to instantiate many tables with an alias, `make_tables` shortcut function has been modified to support tuples with two elements which the first represents the `_table_name` and the second is the `alias`.

### DIFF
--- a/pypika/tests/test_tables.py
+++ b/pypika/tests/test_tables.py
@@ -3,8 +3,8 @@ import unittest
 from pypika import (
     Schema,
     Table,
+    Tables
 )
-from pypika.queries import make_tables
 
 __author__ = "Timothy Heys"
 __email__ = "theys@kayak.com"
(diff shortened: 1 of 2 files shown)


pypika/tests/test_tables.py
```diff
@@ -4,6 +4,7 @@
     Schema,
     Table,
 )
+from pypika.queries import make_tables
```
twheys: this function is given an alias in `pypika/__init__.py` as `Tables`. Could you import it using that, please?

`from pypika import Tables`

PinaGamer: Sure. No problem. 

As you notice there is no any restriction creating many tables at the same time (i.e use the same alias only once or create many tables with the same name). I've supposed that pypika does not restrict anything about that.

PR #269, opened 2019-07-10
Remove 3rd party dependency and drop Python 2 support

teror4uks:
Deprecation Warning still getting on the latest version aenum (2.1.2).  I thought, maybe it's time to drop support Python 2 and remove 3d party dependency `aenum`? `Enum` in standart library Python 3 fully coverage all, what it needed thous library.

`DeprecationWarning: inspect.getargspec() is deprecated, use inspect.signature() or inspect.getfullargspec()
  args, varargs, keywords, defaults = inspect.getargspec(method)`


pypika/__init__.py
```diff
@@ -83,4 +83,4 @@
 
 __author__ = 'Timothy Heys'
 __email__ = 'theys@kayak.com'
-__version__ = '0.28.1'
+__version__ = '0.29.0'
```
twheys: Could you please revert the version increment? Otherwise we run the risk of conflicts if something else gets merged ahead of this PR. I'll set this on the master branch once merged.

twheys:
Thanks for the PR. This is something I've been meaning to do for a little while now.

pypika/__init__.py
twheys: Actually, nevermind that, I'll just merge now so no problem.

PR #281, opened 2019-07-31
Added ability to have fieldless ON CONFLICT DO NOTHING

tomnieuwland:
While using the PostgreSQL query builder in PyPika recently, I discovered there was no way to perform a ON CONFLICT DO NOTHING, with PyPika only allowing for ON CONFLICT (abc.xyz) DO NOTHING queries. 

ON CONFLICT DO NOTHING is useful in PostgreSQL for when you do not know which column is going to conflict, however in the event that any column conflicts, DO NOTHING should be applied. 

There is no need to implement ON CONFLICT DO UPDATE functionality, as PostgreSQL requires that ON CONFLICT is given a field to perform a DO UPDATE

### DIFF
--- a/pypika/dialects.py
+++ b/pypika/dialects.py
@@ -180,22 +180,28 @@ class PostgreQueryBuilder(QueryBuilder):
             return Field(term, table=self._insert_table)
 
     def _on_conflict_sql(self, **kwargs):
-        if self._on_conflict_field is None:
-            return ''
-        elif not self._on_conflict_do_nothing and len(self._on_conflict_updates) == 0:
-            raise QueryException('No handler defined for on conflict')
+        if not self._on_conflict_do_nothing and len(self._on_conflict_updates) == 0:
+            if not self._on_conflict_field:
+                return ''
+            else:
+                raise QueryException('No handler defined for on conflict')
         else:
-            conflict_query = ' ON CONFLICT (' + self._on_conflict_field.get_sql(with_alias=True, **kwargs) + ')'
+            conflict_query = ' ON CONFLICT'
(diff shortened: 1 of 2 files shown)


pypika/dialects.py
```diff
@@ -180,22 +180,28 @@ def _conflict_field_str(self, term):
...
+                conflict_query += ' (' + self._on_conflict_field.get_sql(with_alias=True, **kwargs) + ')'
             if self._on_conflict_do_nothing:
                 conflict_query += ' DO NOTHING'
             elif len(self._on_conflict_updates) > 0:
-                conflict_query += ' DO UPDATE SET {updates}'.format(
-                    updates=','.join(
-                        '{field}={value}'.format(
-                            field=field.get_sql(**kwargs),
-                            value=value.get_sql(**kwargs)) for field, value in self._on_conflict_updates
+                if self._on_conflict_field:
```
twheys: can you please try to reduce the number of nested if statements in this function?

pypika/tests/test_inserts.py
```diff
@@ -160,6 +160,14 @@ def test_non_insert_on_conflict_do_update(self):
             query = PostgreSQLQuery.update(self.table_abc).set('foo', 'bar').on_conflict(
                   'id').do_update(['name'], ["m"])
 
+    def test_insert_on_fieldless_conflict_do_nothing(self):
+        query = PostgreSQLQuery.into(self.table_abc).insert(1).on_conflict(None).do_nothing()
```
twheys: Maybe it would be more obvious to not pass an argument to `on_conflict` instead of an explicit `None`? What do you think?

PR #290, opened 2019-08-07
Support collate on

kongzii:
Support COLLATE when joining tables.

Example:
`q1 = Query.from_(table_a).select(table_b.ouch).join(table_b).on(table_a.foo == table_b.boo, collate="utf8_general_ci")`

Creates `SELECT "b"."ouch" FROM "a" JOIN "b" ON "a"."foo"="b"."boo" COLLATE utf8_general_ci`

### DIFF
--- a/pypika/queries.py
+++ b/pypika/queries.py
@@ -1175,14 +1175,14 @@ class JoinOn(Join):
     def __init__(self, item, how, criteria, collate=None):
         super(JoinOn, self).__init__(item, how)
         self.criterion = criteria
-        self.collate = " COLLATE {}".format(collate) if collate else ""
+        self.collate = collate
 
     def get_sql(self, **kwargs):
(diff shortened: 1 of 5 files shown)


pypika/queries.py
```diff
@@ -1172,15 +1172,17 @@ def replace_table(self, current_table, new_table):
 
 
 class JoinOn(Join):
-    def __init__(self, item, how, criteria):
+    def __init__(self, item, how, criteria, collate=None):
         super(JoinOn, self).__init__(item, how)
         self.criterion = criteria
+        self.collate = " COLLATE {}".format(collate) if collate else ""
```
twheys: Would be better to produce the sql output in the `get_sql` function and store the raw values here in case you need access to how the query was produced.

kongzii: Yes you are right. I updated it.

PR #300, opened 2019-08-20
Support simple SQL Server TOP clauses

paul-ko:
This PR implements supports for simple SQL Server `TOP` clauses, which limit the number of rows returned by a query.  Official documentation on the syntax is here: https://docs.microsoft.com/en-us/sql/t-sql/queries/top-transact-sql?view=sql-server-2017.

The basic idea is that after your `SELECT` keyword, you specify `TOP (<number>)`, where `<number>` specifies that max number of rows to retrieve; for example, `SELECT TOP(1) xyz FROM abc ORDER BY xyz DESC`. The parentheses are optional but recommended, so I've included them.

The `PERCENT` and `WITH TIES` arguments aren't supported.  I can take a look at those if that would be helpful/important.

Thanks!


paul-ko:
Thank you!  You've made my day.

Issue #220, opened 2019-02-21
INSERT with subquery doesn't wrap with parentheses


stickystyle:
Using v0.25.1, if I try to do an insert where one of the values is a subquery, pypika doesn't wrap the subquery in parentheses within the `VALUES` section.

```python
purchase_order_item = pypika.Table("purchase_order_item")
part = pypika.Table("part")

subquery = (
            pypika.MySQLQuery.from_(part)
            .select(part.part_id)
            .where(part.part_number == "FOOBAR")
        )

pypika.MySQLQuery.into(purchase_order_item).columns(purchase_order_item.id_part, purchase_order_item.id_customer).insert(subquery, 12345)
```

The above code generates this invalid query...
```sql
INSERT INTO `purchase_order_item` 
(`id_part`,`id_customer`) 
VALUES 
(SELECT `part_id` FROM `part` WHERE `part_number`='FOOBAR',12345)
```

When it should be...
```sql
INSERT INTO `purchase_order_item` 
(`id_part`,`id_customer`) 
VALUES 
((SELECT `part_id` FROM `part` WHERE `part_number`='FOOBAR'),12345)
```

twheys:
Opened a PR with a fix for this

PR #301, opened 2019-08-20
Add support for subqueries in JoinOn criterion

sangaline:
Subqueries aren't currently wrapped in parentheses when used in a `JOIN ON` criterion, similar to the issue reported in #220. This PR specifies `subquery=True` when calling `criterion.get_sql()` so that subqueries do get wrapped as expected. It also includes a corresponding test case in `test_joins.py`.

Note that the Python 3.4 tests are failing, but this is unrelated to these changes.


twheys:
Thanks for the fix!

PR #306, opened 2019-08-22
Added JSON data type class an refactored JSON operators

### LINKED ISSUE #277: JSONField doesn't respect table kwargs
```
json_field = JSONField(val="test_column", table="test_table", alias=None)
print(json_field.get_sql(dialect="postgressql"))
```

Yields:
`"test_column"`

It doesn't seem to follow the same approach `Field`.`get_sql()` takes.

twheys:
Fixes #277

Did some refactoring to the implementation of the JSON data type for psql and redshift to align it closer to pypika's style

@teror4uks Perhaps you can review this? Sorry for not giving better guidance on your original PR. There were a couple bits of code that needed to get moved around to keep the JSON and JSON comparators aligned with the rest of the pieces.

I introduced a JSON object for carrying JSON data, which can be used directly, but also Fields can be treated as such. See the unit tests for it.

### DIFF
--- a/pypika/__init__.py
+++ b/pypika/__init__.py
@@ -66,6 +66,7 @@ from pypika.terms import (
     EmptyCriterion,
     Field,
     Interval,
+    JSON,
     Not,
     NullValue,
     Parameter,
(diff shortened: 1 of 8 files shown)


teror4uks:
code looks nice, sorry I didn't see mentions from you in my notifications :(

PR #307, opened 2019-08-23
clickhouse IfNull added

d-ganchar:


pypika/clickhouse/nullable_arg.py
```diff
@@ -0,0 +1,28 @@
+from pypika import Field
+from pypika.terms import Function
+
+
+class IfNull(Function):
+    def __init__(self, field, alt, alias: str = None, schema: str = None):
```
twheys: if you change this to 
```
 def __init__(self, *args, **kwargs):
     super().__init__('NULLIF', name, alias=alias)
```

then you wouldn't need to write a custom implementation for `get_sql`

twheys: if you want to enforce the args, you could do it like
```
 def __init__(self, term, alt, **kwargs):
    super().__init__('NULLIF', term, alt, **kwargs)
```

d-ganchar:
> no need for custom `get_sql` implementation, as the function syntax is standard.

Thanks. I'll fix it.

pypika/tests/clickhouse/test_search_string.py
```diff
@@ -1,33 +1,33 @@
...
-            (Like, "like(toString(\"name\"),'twheys')"),
-            (NotLike, "notLike(toString(\"name\"),'twheys')"),
-        )
 
-        for cls, expected in search_types:
-            self.assertEqual(
-                cls(Field('name'), 'twheys'),
-                expected
-            )
+    @parameterized.expand([
```
twheys: btw this library is a really good addition for the types of tests we write in pypika. Appreciate you adding it.

twheys:
Code looks good, just need to fix the build for py3.5

twheys:
nevermind, looks like one of the tests I added is flapping. Going to merge.

d-ganchar:
> nevermind, looks like one of the tests I added is flapping. Going to merge.

https://github.com/kayak/pypika/pull/314#issue-312707452

PR #311, opened 2019-08-29
Reopening reverted PR to refactor JSON operators

twheys:
Accidentally merged https://github.com/kayak/pypika/pull/306 so I reverted it. Reopening this same PR for review.


sl-victormazzeo:
hello, after this modification, the alias parameter is not working anymore. Am I missing something?

```python
> from pypika import Field
> Field('foo', alias='bar').alias is None
True
```

PR #315, opened 2019-08-30
clickhouse condition added

d-ganchar:


pypika/clickhouse/condition.py
```diff
@@ -0,0 +1,42 @@
...
+        self.schema = schema
+        self.args = ()
+        self.name = 'multiIf'
+
+    def get_sql(self, with_alias=False, with_namespace=False, quote_char=None, dialect=None,
+                **kwargs):
+        return '{name}({conditions}){alias}'.format(
+            name=self.name,
+            conditions=','.join(i.get_sql() for i in self._conditions),
+            alias=' ' + self.alias if self.alias else ''
```
twheys: can use `utils.alias_sql(sql, self.alias)` for this to avoid some code repetition

pypika/clickhouse/condition.py
```diff
@@ -0,0 +1,42 @@
...
+        self.alias = alias
+        self.schema = schema
+        self.args = ()
+        self.name = 'multiIf'
+
+    def get_sql(self, with_alias=False, with_namespace=False, quote_char=None, dialect=None,
+                **kwargs):
+        return '{name}({conditions}){alias}'.format(
+            name=self.name,
+            conditions=','.join(i.get_sql() for i in self._conditions),
```
twheys: need to pass down all of the kwargs otherwise you'll end up with some formatting issues with recursive expressions...

```
conditions=','.join(i.get_sql(with_namespace=with_namespace, quote_char=quote_char, dialect=dialect, **kwargs) 
                              for i in self._conditions),
```

pypika/clickhouse/condition.py
```diff
@@ -0,0 +1,42 @@
...
+        """
+        self._conditions = conditions
+        self.alias = alias
+        self.schema = schema
+        self.args = ()
+        self.name = 'multiIf'
+
+    def get_sql(self, with_alias=False, with_namespace=False, quote_char=None, dialect=None,
+                **kwargs):
+        return '{name}({conditions}){alias}'.format(
```
twheys: this pattern looks like standard Function syntax ... could probably just extend `Function` and pass all of the args up in `super().__init__(*conditions, alias=alias)` and get the same results

pypika/clickhouse/condition.py
```diff
@@ -0,0 +1,42 @@
...
+
+class MultiIf(Function):
+    def __init__(self, conditions: tuple, alias: str = None, schema: str = None):
+        """
+        :param conditions: condition1, then1, condition2, then2...else
+        """
+        self._conditions = conditions
+        self.alias = alias
+        self.schema = schema
+        self.args = ()
```
twheys: Doesn't look like args or schema are used again?

d-ganchar: This is really don't need. But I remember that I got in some case(earlier):

> AttributeError: 'MyCustomFunc' object has no attribute args

At the moment all works fine. Thank you.

pypika/clickhouse/condition.py
d-ganchar: All comments are fixed  except this. I'll do it if you want.
I just want to say a few words about _*args_, _**kwargs_. Many people disagree with me. So it's just my experience and my vision and nothing more.
In some cases using _args_, _kwargs_ is really beautiful and useful. For example _decorators_, _wrappers_ or something like this:

```
def example(*names):
    pass
def example2(name: str, *patterns):
    pass

@dataclass
class ExampleClass:
   # props...

ExampleClass(**props)
```

Maybe using _args_ / _kwargs_ is very **_PYTHONIC_** way. But overuse can make a project just unsupported, hard to understand / debug. Any function / method can turn into a 'black box' or magic. Let's imagine you are opening a new project and you see the class:

```
class Resources:
      def __init__(cpu: int, mem: int):
             pass
```

All is clear and you know how work with it. You can just open class and fast check if you forgot something. Now let's make a few changes(just imagine that you see it for the first time or second):

```
# this is not quite obvious
# Resources(1, 2, disk=100) ? Resources(1, 2, hard_drive=100) ? Resources(1, 2, usage=100)
class Resources:
      def __init__(cpu: int, mem: int, **kwargs):
             pass

# let's add more magic
class Resources(BaseResources):
       def __init__(*args, **kwargs):
             # just some code, calls etc...

# how to work with it?
# Resources(stones=5) ? Resources(water=True) etc.
# and now let's imagine that we have a lot of such classes and they are changing...
```

By the way a few days ago I started to work with **python-novaclient**. And look  [at this](https://github.com/openstack/python-novaclient/blob/master/novaclient/client.py#L234). What parameters will be overridden from **env**? What keys can I use in _kwargs_? What interface returns? Yes, I can read docs or do debugging or something else, but I wouldn't waste time if it was more obvious(I mean just **cmd+p** in _PyCharm_ and I see everything).

Anyway it's just my thoughts. I'll change it if you want. :heart:

twheys: Yeah, I understand and I agree with you that it's in a lot of cases better to explicitly list out arguments so that things like IDEs can give proper hints on what arguments should be passed in.

In this case, the biggest concern would be the API exposed to the end user, or the programmer using pypika in their project. When using a function, they would see the `__init__` function arguments, therefore it's a good idea to list of those arguments explicitly so that they would see what is expected. 

On the other hand, with the `get_sql` function, there's a trade off of having to maintain a LOT more code if we need to do it for ever SQL function we implement (obviously there're a lot of them). In that case, since most SQL functions have the same syntax, `func_name(arg0, arg1, etc)`, we don't need to implement `get_sql` for this every time. So here we have a balance between generalizing the code to reduce duplication. while also exposing the arguments explicitly to the user of the Function class we're creating. I think this is a pretty good compromise. 

One other thing worth noting, since pypika build a recursive tree of elements which it then serializes to SQL, there are a couple of arguments which it needs to pass through this hierarchy, such as the quote_char, and if anywhere in the tree one of these arguments gets lost in one of the branches of the tree, then it will lead to incorrect SQL for the children nodes. This can take a bit of time to debug, because it might not be immediately obvious in which class the argument was dropped. The less custom implementations of `get_sql` we can write, the better, because each implementation of it introduces a new place where bugs can hide.

d-ganchar: ok. done ;)

twheys: would you mind updating the other class too, `If` ?

d-ganchar: @twheys fixed. ths!

twheys: thank you :)

PR #330, opened 2019-10-11
add force index hint

Pearman91:
Add FORCE INDEX hint to enhance performance with query optimizer

### DIFF
--- a/pypika/queries.py
+++ b/pypika/queries.py
@@ -440,6 +440,7 @@ class QueryBuilder(Selectable, Term):
 
         self._with = []
         self._selects = []
+        self._force_indexes = []
         self._columns = []
         self._values = []
         self._distinct = False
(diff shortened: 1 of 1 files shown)


pypika/queries.py
```diff
@@ -1055,6 +1069,12 @@ def _from_sql(self, with_namespace=False, **kwargs):
               for clause in self._from
         ))
 
+    def _force_index_sql(self, **kwargs):
+        return ' FORCE INDEX({indexes})'.format(indexes=','.join(
+            index.get_sql(with_alias=True, subquery=True, **kwargs)
```
twheys: just curious why you're explicitly adding `with_alias=True, subquery=True` here? could result in an exception if **kwargs already contains those values.

pypika/queries.py
```diff
@@ -624,6 +625,16 @@ def replace(self, *terms):
...
         self._replace = True
 
+    @builder
+    def force_index(self, *terms):
+        for term in terms:
+            if isinstance(term, Field):
+                self._force_indexes.append(term)
+            elif isinstance(term, str):
+                self._force_indexes.append(Field(term, table=self._from[0]))
+        if len(self._force_indexes) == 0:
```
twheys: this if statement is stateful, which might lead to undesirable results, if `force_index` is called twice on a query, for example. Maybe you want to validate it so that `force_index` can only be called once and requires at least one argument? You could also syntactically force this by accepting args like `def force_index(self, term, *terms)`, that way python would require always at least one argument

Pearman91:
Moved to  'Add force index #334'

PR #331, opened 2019-10-15
Add copy/load from csv for mysql and vertica

RobinPapke:
Add query syntax to mysql and vertica for loading csv files into tables

### DIFF
--- a/pypika/dialects.py
+++ b/pypika/dialects.py
@@ -92,6 +92,41 @@ class MySQLQueryBuilder(QueryBuilder):
         )
 
 
+class MySQLLoadQueryBuilder:
+    def __init__(self):
+        self._load_file = None
+        self._into_table = None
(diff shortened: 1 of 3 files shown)


pypika/dialects.py
```diff
@@ -79,6 +88,23 @@ def modifier(self, value):
         """
         self._modifiers.append(value)
 
+    @builder
+    def load(self, fp):
+        self._load_file = fp
```
twheys: does it make sense to have this on the query instance? There isn't a corresponding test case for it. Do you intend for load to be the first function called to create the query and for into to always follow, or is there some case where a query might be built using other functions first then load is called?

RobinPapke: We can limit it to be being the first function called for now and into having to follow. I'll update it

PR #334, opened 2019-10-17
Add force index

Pearman91:
Add FORCE INDEX hint to enhance performance with query optimizer

This replaces PR 'add force index hint #330' (I made the changes on branch in my fork - whereas #330 PR was done on patch, thus this new PR)

I added class for Index (previously, I used Field) and tests. Requests and comments from #330 were taken in account.

### DIFF
--- a/pypika/terms.py
+++ b/pypika/terms.py
@@ -483,9 +483,6 @@ class Index(Term):
         super(Index, self).__init__(alias)
         self.name = name
 
-    def fields(self):
-        return None
-
     def get_sql(self, quote_char=None, secondary_quote_char="'", **kwargs):
         return format_quotes(self.name, quote_char)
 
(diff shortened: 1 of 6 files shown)


pypika/terms.py
```diff
@@ -478,6 +478,15 @@ def get_sql(self, with_alias=False, with_namespace=False, quote_char=None, secon
         return field_sql
 
 
+class Index(Term):
+    def __init__(self, name, alias=None):
+        super(Index, self).__init__(alias)
+        self.name = name
+
+    def get_sql(self, quote_char=None, secondary_quote_char="'", **kwargs):
```
twheys: The arg `secondary_quote_char` appears to be unused here. Could probably just omit it.

Pearman91: True. I removed that one.

twheys:
merged and published `0.35.12`

Issue #335, opened 2019-10-18
SELECT with subquery doesn't wrap with parantheses


Kulu-M:
This errror sound really similar to [this one](https://github.com/kayak/pypika/issues/220).

This pypika:
```python
uniques = Query.from_(table).select(*)
query = Query.from_(table).select(table.ID, table.DB_SIZE).where(table.ID).isin(uniques)
```

Generates this SQL

```SQL
SELECT "ID","DB_SIZE" FROM "xxx" 
WHERE "ID" IN
SELECT * FROM yyy
```

According to the documentation, this should work but
--> The parantheses around the inner SELECT are missing and therefore the SQL is invalid.

It should be 

```SQL
SELECT "ID","DB_SIZE" FROM "xxx" 
WHERE "ID" IN
(SELECT * FROM yyy)
```

Kulu-M:
This can be fixed when changing line 

```python
return "{term} {not_}IN {container}".format(
```
To the following:
```python
return '{term} {not_}IN ({container})'.format(
```
in **terms.py** in **class ContainsCriterion** in **def get_sql(self, **kwargs)**

Kulu-M:
Fixed

PR #336, opened 2019-10-18
Added missing parentheses

Kulu-M:
See this issue:

https://github.com/kayak/pypika/issues/335

### DIFF
--- a/pypika/terms.py
+++ b/pypika/terms.py
@@ -708,9 +708,8 @@ class ContainsCriterion(Criterion):
     def fields(self):
         return self.term.fields() if self.term.fields else []
 
-    def get_sql(self, **kwargs):
-        # FIXME escape
-        return '{term} {not_}IN {container}'.format(
+    def get_sql(self, subquery=None, **kwargs):
+        return "{term} {not_}IN {container}".format(
               term=self.term.get_sql(**kwargs),
               container=self.container.get_sql(subquery=True, **kwargs),
               not_='NOT ' if self._is_negated else ''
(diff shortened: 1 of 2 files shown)


Kulu-M:
Ok I see this broke a lot of stuff. Maybe some of you experts on pypika have another idea.

twheys:
Can you please provide a test case for this? The change looks incorrect to me, since if this refers to a database column the parentheses would not be required, for example `SELECT * FROM table WHERE 'a' in foo`

In other cases, where a tuple is used, the python tuple should end up being wrapped by the class `pypika.Tuple`. Is there a case where that is not occurring?

Kulu-M:
> In other cases, where a tuple is used, the python tuple should end up being wrapped by the class `pypika.Tuple`. Is there a case where that is not occurring?

In the [issue](https://github.com/kayak/pypika/issues/335), I provided an example where the pypika is not generating a valid SQL query.

twheys:
Ok, I understand. In that case, instead of adding the parenthesis in the sql, you should probably set kwarg `subquery=True`, because this change could lead to cases where parentheses are added where they shouldn't be. In the example below the subquery argument is explicitly set to true for the container argument. You might not need to remove it from kwargs here, can try it without doing that and see if the tests pass. I think this should fix it for your case, but let me know if it doesn't.

Please also add a test to cover your use case.

```
    def get_sql(self, subquery=None, **kwargs):
        return "{term} {not_}IN {container}".format(
              term=self.term.get_sql(**kwargs),
              container=self.container.get_sql(subquery=True, **kwargs),
              not_='NOT ' if self._is_negated else ''
        )
```

Kulu-M:
@twheys I added the test case, everything is working now.

twheys:
Looks good. Merged and published a new patch version `0.35.11`

PR #338, opened 2019-10-18
Allow raw sql string usage

### LINKED ISSUE #337: Compensate missing functions or dialects with raw sql
We're starting to use PyPika with Redshift in my work, and we're facing some troubles as I already saw some people facing here.

The problem is that as PyPika tries to be a generic query builder (as it should be), but as we're using it with Redshift, we're facing some differences in function-wise. 

E.g.:

```
Query.from_('table').select(CurDate())
```
will return
```
SELECT CURRENT_DATE() FROM table
```

Redshift doesn't understand this CURRENT_DATE(). The solution that we're using now extends `Function` and create the `RawFunction`, a function that we can do something like this:

```
Query.from_('table').select(RawFunction('current_date'))
```
and will return
```
SELECT current_date FROM table
```

I understand that with it we can open margin for errors, but also we can give to the user more flexibility to the reserved functions for each type of database.

I'm opening a PR with my solution. I hope that we can discuss it to get into an agreement.

Also, I'm trying to implement a way to use the same function for different dialects, respecting its syntax. But I think it's a faster way to solve some problems.

airtonzanon:
As some dialects don't understand the current functions, it's a way to
use some reserved functions in all dialects without the need to change
the core for each.

I understand that with it we can open a margin for errors, but also we
can give to the user more flexibility.

Closes #337

### DIFF
--- a/pypika/tests/test_functions.py
+++ b/pypika/tests/test_functions.py
@@ -405,6 +405,11 @@ class StringTests(unittest.TestCase):
 
         self.assertEqual("SELECT current_date FROM \"abc\"", str(q))
 
+    def test__raw__sql_on_select_with_alias(self):
+        q = Q.from_(self.t).select(fn.Raw('current_date', 'now'))
+
+        self.assertEqual("SELECT current_date \"now\" FROM \"abc\"", str(q))
+
     def test__raw__sql_on_where(self):
         q = Q.from_(self.t).select('foo').where(fn.Raw('foo != bar'))
 
--- a/pypika/tests/test_functions.py
+++ b/pypika/tests/test_functions.py
@@ -400,6 +400,16 @@ class StringTests(unittest.TestCase):
 
         self.assertEqual("SELECT SUBSTRING(\"foo\",2,6) FROM \"abc\"", str(q))
 
+    def test__raw__sql_on_select(self):
+        q = Q.from_(self.t).select(fn.Raw('current_date'))
+
+        self.assertEqual("SELECT current_date FROM \"abc\"", str(q))
+
+    def test__raw__sql_on_where(self):
+        q = Q.from_(self.t).select('foo').where(fn.Raw('foo != bar'))
+
+        self.assertEqual("SELECT \"foo\" FROM \"abc\" WHERE foo != bar", str(q))
+
 
 class SplitPartFunctionTests(unittest.TestCase):
     t = T('abc')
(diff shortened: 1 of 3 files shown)


twheys:
Thanks for the PR. I would like to suggest, however, that this get presented a little bit differently. The functions provided by pypika are not meant to be an exhaustive list, given that they can be very different from one database provider to the next. Instead, the intended use was to override the `pypika.Function` class. However, I've been thinking for some time that it would make sense to provide a factory function for this instead of extending the class, so that a function can be created with a one liner. This would work for most cases, except for when the function syntax deviates from `name(arg1, arg2, ... argn)`.

Would you like to adapt your PR to implement this instead? The function factory should take a function name argument and a list argument for argument names and return a callable that can be used in the query builder. Here's an example:

```
DateDiff = pypika.function('DATE_DIFF', ['interval', 'start_date', 'end_date'])

query = Query.from_(my_table).select(DateDiff('day', my_table.start_date, my_table.end_date))
print(str(query))
// --> SELECT DATE_DIFF('day', start_date, end_date) FROM my_table
```

pypika/tests/test_functions.py
```diff
@@ -400,6 +400,21 @@ def test__substring(self):
 
         self.assertEqual("SELECT SUBSTRING(\"foo\",2,6) FROM \"abc\"", str(q))
 
+    def test__raw__sql_on_select(self):
+        q = Q.from_(self.t).select(fn.Raw('current_date'))
+
+        self.assertEqual("SELECT current_date FROM \"abc\"", str(q))
```
twheys: For this syntax, where there are no parentheses, you should use the `PseudoColumn` class.

```
CurrentDate = pypika.PseudoColumn('current_date')
Q.from_(self.t).select(CurrentDate)
# --> SELECT current_date FROM "abc"
```

airtonzanon:
Thanks for the review, I really liked the way that you proposed. I'll implement that.

pypika/tests/test_functions.py
airtonzanon: Uhh, nice think, we don't have this in the documentation. May I create a note about this there?

airtonzanon:
I'll close this PR and submit the other implementation in a new one.

pypika/tests/test_functions.py
twheys: yeah there are a few minor features that are missed in the docs. If you would like to add something in there, it would most likely be very helpful for the next person who needs it.

twheys:
Thanks, appreciate the contributions :)

PR #343, opened 2019-10-21
Custom Function implementation

### LINKED ISSUE #337: Compensate missing functions or dialects with raw sql

airtonzanon:
As some dialects don't understand the current functions, it's a way to
use some reserved functions in all dialects without the need to change
the core for each.

This works as a Factory, it gets the function name and arguments and
return a callable that can be used in the query builder.

It also works for functions that don't need arguments.

It was discussed on PR #338 
It closes #337

### DIFF
--- a/pypika/terms.py
+++ b/pypika/terms.py
@@ -19,6 +19,7 @@ from pypika.enums import (
 )
 from pypika.utils import (
     CaseException,
+    FunctionException,
     builder,
     format_alias_sql,
     format_quotes,
@@ -1067,6 +1068,31 @@ class Not(Criterion):
     @property
     def tables_(self):
         return self.term.tables_
+
+
+class CustomFunction():
+    def __init__(self, name, params=None):
+        self.name = name
+        self.params = params
+
+    def __call__(self, *args, **kwargs):
+        if not self._has_params():
+            return Function(self.name, alias=kwargs.get('alias'))
+
+        if not self._is_valid_function_call(*args):
+            raise FunctionException("Function {name} require these arguments ({params}), ({args}) passed".format(
+                                        name=self.name,
+                                        params=', '.join(str(p) for p in self.params),
+                                        args=', '.join(str(p) for p in args)
+                                    ))
+
+        return Function(self.name, *args, alias=kwargs.get('alias'))
+
+    def _has_params(self):
+        return self.params is not None
+
+    def _is_valid_function_call(self, *args):
+        return len(args) == len(self.params)
 
 
 class Function(Criterion):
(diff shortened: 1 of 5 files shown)


pypika/terms.py
```diff
@@ -1060,6 +1061,25 @@ def tables_(self):
         return self.term.tables_
 
 
+class CustomFunction():
+    def __init__(self, term, params=None):
+        self.term = term
```
twheys: Would rename term to `name`, since the verbatim term is already used to refer to objects that represent parts of SQL expressions

pypika/terms.py
```diff
@@ -1060,6 +1061,25 @@ def tables_(self):
...
+        self.params = params
+
+    def __call__(self, *args, **kwargs):
+        if (self.params is None):
```
twheys: this is minor, but we generally omit the params in if statements in other parts of the code

pypika/terms.py
airtonzanon: Nice catch. I'll change it

pypika/terms.py
airtonzanon: I don't think that I got this one. What do you mean?

airtonzanon: Can you review if was it that you meant?

pypika/terms.py
```diff
@@ -1060,6 +1061,31 @@ def tables_(self):
...
+class CustomFunction():
+    def __init__(self, name, params=None):
+        self.name = name
+        self.params = params
+
+    def __call__(self, *args, **kwargs):
+        if not self.hasParams():
+            return Function(self.name, alias=kwargs.get('alias'))
+
+        if not self.isValidFunctionCall(*args):
```
twheys: could you please use slug case for function names? That's the pep8 standard.

twheys: should also prefix the names with an underscore `_` to indicate that they are not meant to be called from outside the class.. `_is_valid_function_call` and `_has_params`

airtonzanon: Done

twheys:
thanks! merged and published as `0.35.13`

PR #361, opened 2019-12-13
Poetry + Black

twheys:
- Introduced poetry for dependency management. Replaced the requirements files with a pyproject.toml file

- added black formatter
- added commit hooks which run black on a pre-commit hook
- added a contributing file which explains how to setup the project and open pull requests


.pre-commit-config.yaml
```diff
@@ -0,0 +1,15 @@
+repos:
+  - repo: https://github.com/ambv/black
+    rev: stable
+    hooks:
+      - id: black
+        language_version: python3.6
+
+  - repo: local
```
mikeengland: You can remove this eslint setup

twheys: oh yeah :)

mikeengland:
@twheys as contributors are starting to use Black it would be worth getting this in soon, especially with the useful contribution guide!

As we now use GitHub actions we now need to do the following:

- [ ] Remove eslint as commented above
- [ ] Add new dependencies to the pyproject.toml file's [tool.poetry.dev-dependencies] section that were added to requirements-dev.txt 
- [ ] Uncomment the Black check in .github/workflows/pythonpackage.yml (so build fails if it doesn't abide by Black's style)
- [ ] Change the 'Install dependencies' job in .github/workflows/pythonpackage.yml to install from Poetry instead of requirements-dev.txt
- [ ] Change tox.ini to install from Poetry and not from requirements-dev.txt.

I _think_ that is everything to get the build working!

taybin:
Why was this closed without being merged after it was approved?

gl3nn:
Black was introduced in a different PR: https://github.com/kayak/pypika/pull/493. Is there something else in this PR that should be introduced?

taybin:
@gl3nn no, I was just curious, since you had approved the PR before.  Thanks for the answer!

PR #364, opened 2019-12-27
Added nodes_ and find_ functions to search the object graph

twheys:
Replaced some utility functions which look for entities inside a pypika query object graph with a more generalized nodes_ and find which filters object nodes based on object type

### DIFF
--- a/pypika/queries.py
+++ b/pypika/queries.py
@@ -11,6 +11,7 @@ from pypika.terms import (
     Field,
     Function,
     Index,
+    Node,
     Rollup,
     Star,
     Term,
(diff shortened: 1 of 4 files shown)


pypika/queries.py
```diff
@@ -889,12 +898,16 @@ def __add__(self, other):
     def __mul__(self, other):
         return self.union_all(other)
 
+    @builder
+    def slice(self, slice):
+        self._offset = slice.start
+        self._limit = slice.stop
```
gl3nn: Shouldn't the limit be `stop-start`? I could be wrong of course I didn't look into the caller or the usage.

gl3nn: I guess since it worked like that before beneath, I'm sure it's still correct.

twheys: didn't think about it, was just moving some old code around. Sounds like it's worth looking into.

PR #374, opened 2020-02-11
Update and Insert improving

nibrag:


pypika/dialects.py
```diff
@@ -328,33 +340,40 @@ def do_update(self, update_field, update_value):
...
+        return super()._distinct_sql(**kwargs)
+
     def _conflict_field_str(self, term):
         if self._insert_table:
             return Field(term, table=self._insert_table)
 
     def _on_conflict_sql(self, **kwargs):
         if not self._on_conflict_do_nothing and len(self._on_conflict_updates) == 0:
-            if not self._on_conflict_field:
+            if not self._on_conflict_fields:
```
twheys: the logic in this function is really complex. Looks like it was already complex even before you change, but would you mind simplifying it a bit to make it easier to read?

There are a couple of places where the indentation is redundant. For example, this if branch returns, so the else statement is redundant. The second else is also redundant because every branch of the if statement returns or raises an exception. I think getting rid of these redundant branches will make the code easier to follow because there will be less branching.

pypika/tests/dialects/test_postgresql.py
```diff
@@ -207,3 +207,14 @@ def test_json_has_any_keys(self):
         self.assertEqual(
             "SELECT * FROM \"abc\" WHERE \"json\"?|ARRAY['dates','imported']", str(q)
         )
+
+
+class DistinctONTests(unittest.TestCase):
```
twheys: very minor, but noticed so will point it out, the N in On here should perhaps be lowercased.

PR #386, opened 2020-03-23
Use self._wrapper_cls in QueryBuilder.set

jimorie:
This commit fixes a problem where the QueryBuilder.set method always
used the ValueWrapper class instead of the instance's _wrapper_cls.

Signed-off-by: Petter Nyström <jimorie@gmail.com>

### DIFF
--- a/pypika/queries.py
+++ b/pypika/queries.py
@@ -901,7 +901,7 @@ class QueryBuilder(Selectable, Term):
     @builder
     def set(self, field, value):
         field = Field(field) if not isinstance(field, Field) else field
-        self._updates.append((field, ValueWrapper(value)))
+        self._updates.append((field, self._wrapper_cls(value)))
 
     def __add__(self, other):
         return self.union(other)
--- a/pypika/tests/test_updates.py
+++ b/pypika/tests/test_updates.py
@@ -1,6 +1,6 @@
 import unittest
 
-from pypika import Table, Query, PostgreSQLQuery, AliasedQuery
+from pypika import Table, Query, PostgreSQLQuery, AliasedQuery, SQLLiteQuery
 
 __author__ = "Timothy Heys"
 __email__ = "theys@kayak.com"
@@ -132,3 +132,12 @@ class PostgresUpdateTests(unittest.TestCase):
         self.assertEqual(
             'UPDATE "abc" SET "lname"="bcd"."long_name" FROM "bcd" RETURNING "abc"."id","bcd"."fname"', str(q)
         )
+
+
+class SQLLiteUpdateTests(unittest.TestCase):
+    table_abc = Table("abc")
+
+    def test_update_with_bool(self):
+        q = SQLLiteQuery.update(self.table_abc).set(self.table_abc.foo, True)
+
+        self.assertEqual('UPDATE "abc" SET "foo"=1', str(q))


jimorie:
I have realized there are a lot of other places where the QueryBuilder's `_wrapper_cls` isn't used, so the fix in this PR is incomplete.

```
>>> t1 = pypika.Table("abc")
>>> t2 = pypika.Table("def")
>>> pypika.SQLLiteQuery.from_(t1).select(t1.id, True).where(t1.foo == True).join(t2).on(t2.bar == True)
SELECT "abc"."id",1 FROM "abc" JOIN "def" ON "def"."bar"=true WHERE "abc"."foo"=true
>>> pypika.SQLLiteQuery.into(t1).columns(t1.id, t1.foo).insert(1, True)
INSERT INTO "abc" ("id","foo") VALUES (1,true)
```

I kind of expected SQLiteQuery's custom ValueWrapper to turn *all* instances of `True` into `1` in the queries above. As it stands, it only seems to work in the select part.

Is this working as intended?

jimorie:
I'm not really bothered about the use in the MySQL dialect code -- I assume it's not a problem there. But it seems to be a problem that the `_wrapper_cls` isn't used consistently in the core code.

I guess the problem is in `Term.wrap_constant()`. That method takes an optional `_wrapper_cls` argument, but that argument is only provided in *one* place (except dialects), and that is in the select code. If not given, it defaults to a `ValueWrapper`.

It looks like a lot of work to fix this, unfortunately. My first instict is that all `Term` objects need to be initialized with the wrapper class it should use by default. But there are a lot of places where new Terms are created...

Would you consider a change that introduced `_wrapper_cls` as an attribute on `Term` objects? Given all the added complexity and (possibly) overhead such a change would mean? I could perhaps look at it, but only if you think there is a possibility for it to be merged.

Do you have any other ideas how to make the SQLLite dialect's use of a customer `ValueWrapper` (and similar use cases) consistent?

twheys:
Ah I see what you mean, definitely an oversight. I think the implementation of this `_wrapper_cls` function was just a quick fix for MSSQL for this case with booleans. Maybe a quicker and more effective solution would be to get rid of the `SQLLiteValueWrapper` class and move the boolean clause into the `ValueWrapper` class directly. There are some other dialect specific cases handled in the main class, so it wouldn't deviate terribly, and given this is so far the only corner case with ValueWrapper, I think it can be viewed as a corner case and treated as such.

Something like this:
```
class ValueWrapper(Term):
    ...
    def get_value_sql(self, **kwargs):
        dialect = kwargs.get("dialect", None)
        ...
        if dialect == Dialects.MSSQL and isinstance(self.value, bool):
            return "1" if self.value else "0"
        if isinstance(self.value, bool):
            return str.lower(str(self.value))
        ...
```

Then we could just get rid of the `_wrap_constant` function entirely.

jimorie:
Well, as a user of the library I definitely appreciate that I can write my own dialect and have it able to manipulate how it formats individual terms in the output, or attach side effects to them. I can see this being provided either by a configurable `ValueWrapper` as currently (although then it would have to be used consistently), or perhaps by adding some sort of optional converter parameter to the `get_sql` methods. It sounds as if this would not be as easy with what you're suggesting?

I can understand if you deem this as out of scope for what the library should offer, though. 

For the record, my actual use case here was to write my own dialect that automatically turned all values used in the query into `Parameter` instances, and adding their values to a args dict on the query builder. So that I would get a parameterized sql query and have the arguments collected and ready to go. :)

twheys:
I guess either way you would need to override the main Query class to make a dialect. There you could then override the _wrap_constant function, instead of providing a separate version of ValueWrapper for each dialect. Looks like the wrapper_cls might be inaccessible in a lot of places where it might be necessary to use, so it would certainly add complexity passing that around to all the different objects in the object tree

jimorie:
Sorry, I forgot about this. Glad to see it merged even though it didn't do enough for my purposes of generating parameterized SQL. I ended up doing some post-processing in the  `QueryBuilder` that replace all instances of `ValueWrapper.value` with instances of `Parameter`. Cheers!

Issue #389, opened 2020-03-26
Interval produces invalid sql for units with 0 values


Kamforka:
# System environment

OS : Ubuntu 16.04 LTS
PyPika : 0.36.0

# Overview

Instantiating an `Interval` object with default 0 values or with empty argument list makes it to output invalid sql.

# Steps to reproduce

```python3
from pypika import Interval
Interval(days=0).get_sql()
```

# Expected behavior

```python
>>> Interval(days=0).get_sql()
"INTERVAL '0 DAY'"
```

# Actual behavior

```python
>>> Interval(days=0).get_sql()  # same happens with Interval().get_sql()
"INTERVAL '0 None'"
```

PR #391, opened 2020-03-30
Set default unit with Interval query

lntuition:
Current `Interval` code can generate wrong query When all value is unset.
`(e.g Interval(days=0).get_sql())`

So just check that all value is unset, and set default with `DAY`. 
This is small fixing, you may don't like this solution. 
Maybe we need to consider as `exception` when all value is unset.
Please check it and review plz. Thank u :)

Fixed : #389

### DIFF
--- a/pypika/terms.py
+++ b/pypika/terms.py
@@ -1425,6 +1425,10 @@ class Interval(Node):
                 else self.largest
             )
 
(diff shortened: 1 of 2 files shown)


twheys:
Good catch! Thanks for the fix :)

Issue #388, opened 2020-03-25
Alias missing from custom class in PyPika


lmanideepkumar:
I am trying to implement custom class **Struct** 
which is not availble in PyPIKA. I am trying to acheive custom class which accepts FIRST function as input parameter.But alias is missing for First() function in the generated output even if we are passing as alias as parameter for First() function.

**Code Snippet:**

from pypika import (Field as F,Interval,Case,JoinException,JoinType,MySQLQuery,Query,CustomFunction,Order,Table,Tables,UnionException,functions as fn)
from pypika.terms import (Function)

class Struct(Function):
  def __init__(self, *args,alias=None):
    super(Struct, self).__init__('STRUCT', *args, alias=alias)        

customers = Table('customers')
query = Query.from_(customers).select(Struct(fn.First("customers.Addres").as_("Addr"),fn.First("customers.Phone").as_("Ph")).as_("ContandPref"))

**generated output:** 
SELECT STRUCT(FIRST('customers.Addres'),FIRST('customers.Phone')) "ContandPref" FROM "customers"

**Expected output:**
SELECT STRUCT(FIRST('customers.Addres') **Addr** ,FIRST('customers.Phone') **Ph**) "ContandPref" FROM "customers"

gl3nn:
https://github.com/kayak/pypika/pull/392 had a fix for this but it had too many side effects. This was reverted in https://github.com/kayak/pypika/pull/433. See the discussions there.

PR #392, opened 2020-04-01
Allow alias with nested function

lntuition:
With #388, when using function nest. Inner alias is ignored.
I don't know that this behavior is intended.
Allowing alias with nested function (inner) can be useful
(Maybe custom function?.. umm.. i'm not sure 😕, Plz see #388)
And add test to test this feature.

Fixed: #388


gl3nn:
I'm going to be undoing the changes in this pull request as this has broken quite a few usages. Having the argument fields of a function display their alias should probably configured in a different way. Maybe a keyword argument for the outer function?

Issue #377, opened 2020-03-07
Fails to install on py39-dev


grigi:
Hi, I tried to install pypika on Python3.9 alpha, and it failed with the following error:
```
Collecting pypika==0.35.21
  Downloading PyPika-0.35.21.tar.gz (84 kB)
    ERROR: Command errored out with exit status 1:
     command: /home/travis/virtualenv/python3.9-dev/bin/python -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-install-wotkbnjo/pypika/setup.py'"'"'; __file__='"'"'/tmp/pip-install-wotkbnjo/pypika/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /tmp/pip-install-wotkbnjo/pypika/pip-egg-info
         cwd: /tmp/pip-install-wotkbnjo/pypika/
    Complete output (7 lines):
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-install-wotkbnjo/pypika/setup.py", line 39, in <module>
        version=version(),
      File "/tmp/pip-install-wotkbnjo/pypika/setup.py", line 13, in version
        with open(path, 'rU') as file:
    ValueError: invalid mode: 'rU'
```

This was in a Travis build job using version `3.9-dev`
The issue is the `U` (universal newlines) have been deprecated since Python3.4: https://docs.python.org/3.4/library/functions.html#open
It may be fine to just drop that option from `open()`.

lntuition:
With current version docs(version 3.8.2) : https://docs.python.org/3/library/functions.html?highlight=open#open

> There is an additional mode character permitted, 'U', which no longer has any effect, and is considered deprecated. It previously enabled universal newlines in text mode, which became the default behaviour in Python 3.0. Refer to the documentation of the newline parameter for further details.

It is no effect and work default behavior in python3. So i will add minor patch that remove 'U' option.

PR #398, opened 2020-04-16
Remove deprecated option in builtin function `open`

lntuition:
- With builtin function open, 'U' option to mode is deprecated
- Python3 support 'U' option with default behavior

Fixed: #377


mikeengland:
Thanks for the contribution. I'll also add 3.8 and 3.9 alpha to tox/travis later on.

PR #406, opened 2020-04-17
fixing gh actions 

mikeengland:
There is another unofficial GH Coveralls action which works with Python. This should fix the issues with the manual Coveralls run which does not work inside PRs.

No need to merge - this PR will be used for testing so may be open for some time.


gl3nn:
Please squash the commits. =)

PR #409, opened 2020-04-17
Add handler for MySQL ON CONFLICT IGNORE statement

droserasprout:
Also added a couple of PostgreSQL ON CONFLICT statement examples to README. Not sure about where to place tests for this feature as many PostgreSQL specific modifier tests are in `test_inserts` module.


mikeengland:
Thanks for the contribution!

PR #421, opened 2020-04-24
Fix tuple table replacement

RobinPapke:


pypika/tests/test_query.py
```diff
@@ -120,6 +115,20 @@ def test_replace_orderby_table(self):
...
+        query = (
+            Query.from_(self.table_a)
+            .select(self.table_a.cost, self.table_a.revenue)
+            .where(Tuple(self.table_a.cost, self.table_a.revenue) == Tuple(1, 2))
+        )
+
+        query = query.replace_table(self.table_a, self.table_b)
+
+        self.assertEqual(
+            "SELECT " '"cost","revenue" ' 'FROM "b" ' 'WHERE ("cost","revenue")=(1,2)',
```
mikeengland, line 128: Is this SQL string correct? It has multiple opening/closing single quotes.

pypika/tests/test_query.py
```diff
@@ -120,6 +115,20 @@ def test_replace_orderby_table(self):
             'SELECT "customer" ' 'FROM "b" ' 'ORDER BY "customer"', str(query)
         )
 
+    def test_replace_tuple_table(self):
+        query = (
+            Query.from_(self.table_a)
+            .select(self.table_a.cost, self.table_a.revenue)
+            .where(Tuple(self.table_a.cost, self.table_a.revenue) == Tuple(1, 2))
```
twheys, line 122: You shouldn't need to use the Tuple class directly. You can just use a python tuple. Pypika wraps python tuples with it's own tuple internally.

pypika/tests/test_query.py
RobinPapke: Fixed the formatting of the sql strings.

pypika/tests/test_query.py
RobinPapke: I've updated it to use a python tuple instead.

PR #425, opened 2020-04-30
Remove alias quotes on Snowflake dialect

### LINKED ISSUE #424: Snowflake ALIAS_QUOTE_CHAR seem to be wrong
Hi all,

we at Joyn have recently started using pypika and we love it. There is only one issue, the SnowFlakeQueryBuilder alias quote char seem to be wrong as it leads to queries that snowflake does not accept. As an example
```python
from pypika.dialects import SnowflakeQuery
from pypika import Table
t = Table("tbl")
q = SnowflakeQuery().from_(t).select(t.name.as_("aliased_name"))
print(q)
### outputs
SELECT name "aliased_name" FROM tbl
```
Expected output would be
```python
SELECT name aliased_name FROM tbl
```
To fix that, we have to set
SnowFlakeQueryBuilder.ALIAS_QUOTE_CHAR = None in every file and project. Not sure if this is something that could be fixed directly in the lib

airtonzanon:
fixes #424

### DIFF
--- a/pypika/dialects.py
+++ b/pypika/dialects.py
@@ -26,7 +26,6 @@ from pypika.utils import (
 
 class SnowFlakeQueryBuilder(QueryBuilder):
     QUOTE_CHAR = None
-    ALIAS_QUOTE_CHAR = '"'
 
     def __init__(self, **kwargs: Any) -> None:
         super().__init__(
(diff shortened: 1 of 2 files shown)


mikeengland:
Thanks for fixing!

Issue #426, opened 2020-05-01
Extract Function not wotking with joins 


abhijo89:
```python
from pypika import Query, Table, DatePart
from pypika.functions import Extract

table1 = Table('table1')
table2 = Table('table2')

query = Query.select(
    table1.id, table2.id, Extract(DatePart.year, table1.created_date)
).from_(table1).join(table2).on(
table1.id == table2.table1_id
)

print(str(query))
```
The query shuld be
```sql
SELECT "table1"."id","table2"."id",EXTRACT(YEAR FROM "table1.created_date") FROM "table1" JOIN "table2" ON "table1"."id"="table2"."table1_id"
``` 
But we get 
```sql
SELECT "table1"."id","table2"."id",EXTRACT(YEAR FROM "created_date") FROM "table1" JOIN "table2" ON "table1"."id"="table2"."table1_id"
```

PR #427, opened 2020-05-03
Add table to Extract function

lntuition:
Add table name to `EXTRACT` function

Fixed : #426

### DIFF
--- a/pypika/functions.py
+++ b/pypika/functions.py
@@ -303,7 +303,7 @@ class Extract(Function):
         self.field = field
 
     def get_special_params_sql(self, **kwargs):
-        return "FROM {field}".format(field=self.field,)
+        return "FROM {field}".format(field=self.field.get_sql(**kwargs))
 
 
 # Null Functions
--- a/pypika/tests/test_functions.py
+++ b/pypika/tests/test_functions.py
@@ -561,6 +561,7 @@ class CastTests(unittest.TestCase):
 class DateFunctionsTests(unittest.TestCase):
     dt = F("dt")
     t = T("abc")
+    t2 = T("efg")
 
     def _test_extract_datepart(self, date_part):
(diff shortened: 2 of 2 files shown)


pypika/functions.py
```diff
@@ -303,7 +303,7 @@ def __init__(self, date_part, field, alias=None):
         self.field = field
 
     def get_special_params_sql(self, **kwargs):
-        return "FROM {field}".format(field=self.field,)
+        return "FROM {table}.{field}".format(table=self.field.table, field=self.field)
```
twheys, line 306: the functionality for this is already built into the field class. The issue here is that the **kwargs are not being passed in which would tell it to render the table name. To remedy that we can call `field.get_sql(**kwargs)` directly on the field object here instead of casting it to a string. That should fix the issue as well but in a way that lets the field object determine it's own serialization logic.

lntuition: Ah. I missed point.. My solution is fool. I fix and add new testcase for join with `EXTRACT`. Thanks for reviewing :smile:

mwilkins91:
This is an issue for all functions, would this solution fix the problem for all of them? (seems like it would, just confirming...)

Also, when could we expect this to be available in the main lib?

lntuition:
@mikeengland Other function do not use `get_special_params_sql` inner. `COALESCE` already works well. See below example plz.

``` python
from pypika import Query, Table
from pypika.functions import Coalesce

t = Table("abc")
t2 = Table("efg")

q = (
    Query.from_(t)
    .join(t2)
    .on(t.id == t2.t_id)
    .select(Coalesce(t.foo, "bar"))
)

print(str(q)) # SELECT COALESCE("abc"."foo",'bar') FROM "abc" JOIN "efg" ON "abc"."id"="efg"."t_id"

```

PR #429, opened 2020-05-04
remove argument duplication

### LINKED ISSUE #428: Postgres: error when RETURNING inside WITH subquery
Hi. I'm trying to execute the following code

```python
from pypika import Table, AliasedQuery, PostgreSQLQuery as Query

customers = Table('customers')
other_table = Table('other')

sub_query = Query.into(customers).columns('a', 'b').insert(1, 2).returning('*')
aliased = AliasedQuery('an_alias')

query = Query.with_(sub_query, 'an_alias').from_(aliased).select(aliased.star)
query = query.left_join(other_table).on(aliased.a == other_table.id)

print(query)
```

and getting an error
```
  File "/home/alex/venv/lib/python3.8/site-packages/pypika/queries.py", line 1080, in __str__
    return self.get_sql(dialect=self.dialect)
  File "/home/alex/venv/lib/python3.8/site-packages/pypika/dialects.py", line 527, in get_sql
    querystring = super(PostgreQueryBuilder, self).get_sql(
  File "/home/alex/venv/lib/python3.8/site-packages/pypika/queries.py", line 1189, in get_sql
    querystring = self._with_sql(**kwargs)
  File "/home/alex/venv/lib/python3.8/site-packages/pypika/queries.py", line 1241, in _with_sql
    return "WITH " + ",".join(
  File "/home/alex/venv/lib/python3.8/site-packages/pypika/queries.py", line 1244, in <genexpr>
    + clause.get_sql(subquery=False, with_alias=False, **kwargs)
  File "/home/alex/venv/lib/python3.8/site-packages/pypika/queries.py", line 79, in get_sql
    return self.query.get_sql(**kwargs)
  File "/home/alex/venv/lib/python3.8/site-packages/pypika/dialects.py", line 538, in get_sql
    querystring += self._returning_sql(with_namespace=with_namespace, **kwargs)
TypeError: _returning_sql() got multiple values for keyword argument 'with_namespace'
```

Dandi91:
Closes #428

### DIFF
--- a/pypika/tests/test_selects.py
+++ b/pypika/tests/test_selects.py
@@ -1187,18 +1187,15 @@ class SubqueryTests(unittest.TestCase):
               str(test_query),
         )
 
-    def test_join_with_with_insert_returning(self):
+    def test_select_from_with_returning(self):
         sub_query = PostgreSQLQuery.into(self.table_abc).insert(1).returning('*')
         test_query = (
             Query.with_(sub_query, "an_alias")
                 .from_(AliasedQuery("an_alias"))
-                .join(self.table_efg)
-                .on(AliasedQuery("an_alias").fizz == self.table_efg.buzz)
                 .select("*")
         )
         self.assertEqual(
-            'WITH an_alias AS (INSERT INTO "abc" VALUES (1) RETURNING *) '
-            'SELECT * FROM an_alias JOIN "efg" ON "an_alias"."fizz"="efg"."buzz"',
+            'WITH an_alias AS (INSERT INTO "abc" VALUES (1) RETURNING *) SELECT * FROM an_alias',
             str(test_query)
         )
 
--- a/pypika/tests/test_selects.py
+++ b/pypika/tests/test_selects.py
@@ -1187,6 +1187,21 @@ class SubqueryTests(unittest.TestCase):
               str(test_query),
         )
 
+    def test_join_with_with_insert_returning(self):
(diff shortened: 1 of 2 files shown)


gl3nn:
Could you add a test for this?

Dandi91:
I wasn't sure whether I should add it in 'select' or 'insert' tests, so feel free to correct me.

pypika/tests/test_selects.py
```diff
@@ -1187,6 +1187,21 @@ def test_join_with_with(self):
               str(test_query),
         )
 
+    def test_join_with_with_insert_returning(self):
```
gl3nn, line 1190: I guess the name should be a bit more descriptive and should also describe the fact that we are testing the alias that was a problem before.

gl3nn: If the part that was broken is the subquery having an alias, then the test should be named like that. Adjust to the actual problem solved accordingly.

Dandi91: It is the `.returning('*')` statement in subquery that is actually broken, not an alias. If we omit it, everything works as expected. If we use it outside of subquery, everything works as expected. Though I'm not sure if it affects only `WITH` queries or inlined subqueries are affected as well (such as `SELECT FROM (INSERT ...)`).

I've simplified the test query a bit and renamed the test, would you mind taking another look at it?

gl3nn: This is perfect, thanks! It's now a lot more clear what the bug is that gets fixed.

Issue #430, opened 2020-05-05
Error with empty argument columns


lntuition:
With support iterable params to columns(issue #372, PR #373). I didn't catch that empty `columns` can be pass to arg of `columns`. Sadly, there is no Testcase that test below cases.

So passing empty argument to `columns` generate errors. Below are simple test code and result.

``` python
from pypika import Query, Table

table = Table("abc")
query = (
    Query.into(table)
    .columns()
    .insert(1, "a", True)
)

print(str(query))
```

Result at 0.35.21
``` SQL
INSERT INTO "abc" VALUES (1,'a',true)
```

Result at 0.37.4
``` bash
  File "a.py", line 5, in <module>
    Query.into(table)
  File "/root/.pyenv/versions/py37/lib/python3.7/site-packages/pypika/utils.py", line 50, in _copy
    result = func(self_copy, *args, **kwargs)
  File "/root/.pyenv/versions/py37/lib/python3.7/site-packages/pypika/queries.py", line 775, in columns
    if isinstance(terms[0], (list, tuple)):
IndexError: tuple index out of range
```

PR #431, opened 2020-05-05
Fix columns with empty args

lntuition:
Fix columns with empty arguments. I occurs by #373 that support iterable params to `columns`
Really sorry for the hasty PR :disappointed_relieved: ...

Fixed: #430


mikeengland:
Thanks for the contribution 👍

PR #433, opened 2020-05-06
Undo the showing of aliases of fields that are arguments of a function

gl3nn:
This is essentially a rollback of https://github.com/kayak/pypika/pull/392. I commented on that pull request as well.

The linked PR made too drastic changes that had a big impact on existing usages. We should either provide a migration guide or find a different solution for this feature.

### DIFF
--- a/pypika/terms.py
+++ b/pypika/terms.py
@@ -1156,7 +1156,7 @@ class Function(Criterion):
         return "{name}({args}{special})".format(
               name=self.name,
               args=",".join(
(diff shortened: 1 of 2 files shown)


gl3nn:
@lntuition

lntuition:
@gl3nn Ah, sure. I didn't catch that change can lead to big waves. I'm agreeing that revert it. Also, colud you reopen #388 when this merged?

PR #439, opened 2020-05-26
Revert #425 remove alias quotes on Snowflake dialect

mikeengland:
This was causing issues when using alias' which contain special characters e.g. $. 

+ @airtonzanon @Shawe82

### DIFF
--- a/pypika/dialects.py
+++ b/pypika/dialects.py
@@ -26,6 +26,7 @@ from pypika.utils import (
 
 class SnowFlakeQueryBuilder(QueryBuilder):
     QUOTE_CHAR = None
+    ALIAS_QUOTE_CHAR = '"'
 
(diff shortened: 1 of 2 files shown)


mikeengland:
@Shawe82 the issue that you may be facing is described here: https://community.snowflake.com/s/question/0D50Z00009eDNPz/select-column-with-double-quotes

If a table was created using quotes, it makes the column names case sensitive.

 

> Identifier Resolution
> 
> By default, Snowflake applies the following rules for storing identifiers (at creation/definition time) and resolving them (in queries and other SQL statements):
> 
>  
> 
> When an identifier is unquoted, it is stored and resolved in uppercase.
> When an identifier is double-quoted, it is stored and resolved exactly as entered, including case.

Issue #250, opened 2019-05-16
Set difference, set intersection support?


roger-rojas:
Is there existing/planned support for set difference or set intersection operators?

e.g. in table "Likes" with columns "Person" and "Pizza":
1) To find people who like _both_ mushroom and onion pizza
SELECT person FROM likes WHERE pizza="mushrooms"
INTERSECT
SELECT person FROM likes WHERE pizza="onions"

2) To find people who like onions _but not_ ham on their pizzas
SELECT person FROM likes WHERE pizza="onions"
MINUS
SELECT person FROM likes WHERE pizza="ham"

I am new to SQL, so maybe there is a similarly simple way to do this, but set operations make these clear.

twheys:
I think this could be implemented in pypika pretty easily, since we have code already for things like unions that produce similar SQL. Would you like to try and raise a PR?

DimpleK26:
@twheys I have raised a PR regarding the above issue. Could you please take a look

roger-rojas:
Wow, thank you @DimpleK26 ! I'm pretty inexperienced with the workflow on here. Now that your changes have been merged, do I go ahead and click "Close Issue"?

DimpleKannadasan:
@roger-rojas Yes . If you are happy with intersect ,minus and except operations, kindly go ahead :)

PR #447, opened 2020-06-16
Changes to include Set Operations

DimpleK26:
- Intersect
- Minus
- Except

- Unit tests for Intersect, Minus, Except Operations
- Fix for existing issues in Union unit tests

https://github.com/kayak/pypika/issues/250

### DIFF
--- a/pypika/enums.py
+++ b/pypika/enums.py
@@ -58,9 +58,12 @@ class JoinType(Enum):
     cross = "CROSS"
 
 
-class UnionType(Enum):
-    distinct = ""
-    all = " ALL"
+class SetOperation(Enum):
+    union = "UNION"
+    union_all = "UNION ALL"
+    intersect = "INTERSECT"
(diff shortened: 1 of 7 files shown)


pypika/enums.py
```diff
@@ -58,9 +58,12 @@ class JoinType(Enum):
     cross = "CROSS"
 
 
-class UnionType(Enum):
-    distinct = ""
-    all = " ALL"
+class SetOperations(Enum):
```
gl3nn, line 61: To be consistent with other Enums it probably makes sense to rename this to the singular SetOperation.

DimpleKannadasan: will do

DimpleKannadasan: @gl3nn Resolved the above comment

DimpleKannadasan:
@gl3nn @twheys Could this Pr be reviewed please. I have resolved the comments

PR #448, opened 2020-06-16
Fix is_aggregate property for Case fields

x8lucas8x:
So Case's criterions were not taken into account when evaluation a Case field's is_aggregate property, which imo is incorrect.

### DIFF
--- a/pypika/tests/test_aggregate.py
+++ b/pypika/tests/test_aggregate.py
@@ -53,6 +53,15 @@ class IsAggregateTests(unittest.TestCase):
         v = 1 / fn.Sum(Field("foo"))
         self.assertTrue(v.is_aggregate)
 
+    def test__agg_case_criterion_is_aggregate(self):
+        v = (
+            Case()
+            .when(fn.Sum(Field("foo")) > 666, 'More than 666')
+            .else_('Less than 666')
+        )
+
+        self.assertTrue(v.is_aggregate)
+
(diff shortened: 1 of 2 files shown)


pypika/tests/test_aggregate.py
```diff
@@ -97,20 +106,15 @@ def test__case_with_field_is_not_aggregate(self):
...
             Case()
             .when(Field("foo") == 1, 1)
             .when(fn.Sum(Field("foo")) == 2, 2)
             .else_(3)
         )
 
-        self.assertFalse(v.is_aggregate)
-
-    def test__case_all_constants_is_aggregate_none(self):
-        v = Case().when(True, 1).when(False, 2).else_(3)
```
mikeengland, line 111: Is this test no longer relevant?

x8lucas8x: I think it makes no sense. It's not testing a real scenario, given the criterion is a constant. That's not how you'd use a CASE statement in SQL.

mikeengland: The use case I can see for it is for renaming values e.g. "true" -> "foo", "false" -> "bar"

x8lucas8x: But then you'd have a conditional. If field x is true then foo.

mikeengland: Makes sense, thanks.

PR #449, opened 2020-06-22
Add implementation of fields_() for EmptyCriterion

gl3nn:
This issue would have also been solved if `EmptyCriterion` was a subclass of `Criterion`, which would make sense but adds a lot of unnecessary overhead.

### DIFF
--- a/pypika/tests/test_criterions.py
+++ b/pypika/tests/test_criterions.py
@@ -763,6 +763,13 @@ class AnyTests(unittest.TestCase):
         self.assertEqual(str(crit), '"a" OR "b" OR "c" OR "d"')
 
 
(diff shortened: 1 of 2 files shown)


pypika/tests/test_criterions.py
```diff
@@ -763,6 +763,13 @@ def test_with_generator(self):
         self.assertEqual(str(crit), '"a" OR "b" OR "c" OR "d"')
 
 
+class EmptyCriterionTests(unittest.TestCase):
+    def test_field_(self):
```
x8lucas8x, line 767: Minor typo in test name. It should be test_fields_ instead.

gl3nn: Good catch.

PR #451, opened 2020-06-23
anti sql injection

kulame:
anti sql injection when use pypika


mikeengland:
Hi @kulame, the best approach is to utilise the built-in implementations of escaping in Python SQL modules e.g. mysqlclient.
To do this, use a Parameter placeholder (https://pypika.readthedocs.io/en/latest/2_tutorial.html#parametrized-queries) and pass in your value as a parameter to the driver.

PR #462, opened 2020-07-03
Fix query quote char

### LINKED ISSUE #446: Snowflake ALIAS Quote chars on joins
@mikeengland Thanks for letting me know that you have reverted the original fix https://github.com/kayak/pypika/pull/439. I now digged a bit deeper into the quote char issues, and it is not in general and issues but only on table joins. To see what I mean, here the following simple example

```python
from pypika import Table
from pypika.dialects import SnowflakeQuery
t1 = Table("data1")
t2 = Table("data2")
q1 = SnowflakeQuery.from_(t1).select(t1.b).as_("q1")
q2 = SnowflakeQuery.from_(t2).select(t2.b).as_("q2")
q3=SnowflakeQuery.from_(q1).join(q2).on(q1.b==q2.b).select("*")
```
would result in
`SELECT * FROM (SELECT b FROM data1) "q1" JOIN (SELECT b FROM data2) "q2" ON q1.b=q2.b ` which doesn't work. What would work is either
`SELECT * FROM (SELECT b FROM data1) q1 JOIN (SELECT b FROM data2) q2 ON q1.b=q2.b` so no quote chars around query aliases, or
`SELECT * FROM (SELECT b FROM data1) "q1" JOIN (SELECT b FROM data2) "q2" ON "q1".b="q2".b` so using the quote chars when referring to the aliases.

Shawe82:
* Add quote chars for queries
Fixes #446

### DIFF
--- a/pypika/dialects.py
+++ b/pypika/dialects.py
@@ -27,6 +27,7 @@ from pypika.utils import (
 class SnowFlakeQueryBuilder(QueryBuilder):
     QUOTE_CHAR = None
     ALIAS_QUOTE_CHAR = '"'
+    QUERY_ALIAS_QUOTE_CHAR = ''
 
     def __init__(self, **kwargs: Any) -> None:
         super().__init__(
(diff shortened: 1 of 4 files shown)


Shawe82:
@mikeengland here I tried to create a fix for the snowflake join quote char issue we've discussed. Have another option in mind if this one doesn't fit.

PR #471, opened 2020-08-04
Fix error: TypeError: get_sql() got multiple values for keyword argument 'groupby_alias'

paobai:
Fix error: TypeError: get_sql() got multiple values for keyword argument 'groupby_alias'

mssql and oracle will raise TypeError when use .get_sql method

set groupby_alias in kwargs ,fix this


mikeengland:
Please can you add some test cases for this change? Thanks!

mikeengland:
I missed this and fixed this in https://github.com/kayak/pypika/pull/530! Thanks anyway!

PR #472, opened 2020-08-06
Added drop statement and improved create statement

nibrag:


mikeengland:
Thanks for the contribution @nibrag. Would you mind adding some documentation for create/drop table as there is currently no documentation?

Aside from that the code looks good.

nibrag:
@mikeengland Added

PR #479, opened 2020-09-04
Ensure bitwiseand function uses wrapped constant value

### LINKED ISSUE #284: Support for bitwise operators
Is it possible to add support for bitwise operators in `WHERE` clause.?
for example if I had the `messages` table with the field `status`  of type INT and the bits of the value of this field represent a specific status i.e _delivered_, _seen_, ..etc. Instead of having a separate BOOL field for each status.

supposing that the first bit is for the _delivered_ status and the second one is for the _seen_ status

If i want to retrieve only delivered messages that are not seen yet, the query would be something like:
**```SELECT `status` FROM `messages` WHERE (`status` & 1 = 1) ```**

mikeengland:
fixes #284

### DIFF
--- a/pypika/terms.py
+++ b/pypika/terms.py
@@ -141,7 +141,7 @@ class Term(Node):
         return self.isnull().negate()
 
     def bitwiseand(self, value: int) -> "BitwiseAndCriterion":
-        return BitwiseAndCriterion(self, value)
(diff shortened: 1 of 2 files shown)


airtonzanon (approved):
LGTM but coveralls failed on finding the build

PR #487, opened 2020-09-19
SnowflakeQuery: You really don't need quote in Snowflake should fix #483 

### LINKED ISSUE #483: Snowflake: alias quote join has still some issues
The unit test is working correctly from #462  :
```
    def test_dont_use_double_quotes_on_joining_queries(self):
        foo = self.table_abc
        bar = self.table_efg
        q1 = SnowflakeQuery.from_(foo).select(foo.b)
        q2 = SnowflakeQuery.from_(bar).select(bar.b)
        q = SnowflakeQuery.from_(q1).join(q2).on(q1.b == q2.b).select("*")

        self.assertEqual(
            "SELECT * "
            'FROM (SELECT b FROM abc) sq0 '
            'JOIN (SELECT b FROM efg) sq1 '
            "ON sq0.b=sq1.b",
            q.get_sql(),
        )
```

However, if you slightly change it to:

```
    def test_dont_use_double_quotes_on_joining_queries(self):
        foo = self.table_abc.as_('foofoo')
        q = SnowflakeQuery.from_(foo).join(foo).on(foo.b == foo.b).select("*")

        self.assertEqual(
            "SELECT * "
            'FROM foo foofoo '
            'JOIN foo foofoo '
            "ON foofoo.b=foofoo.b",
            q.get_sql(),
        )
```
This will fail because the output is:
`SELECT * FROM abc "foofoo" JOIN abc "foofoo" ON foofoo.b=foofoo.b`
I tried on the master branch.

zippeurfou:
I removed the quote from Snowflake.
The query run and it just make everything easier :) 
It should close #483

### DIFF
--- a/pypika/dialects.py
+++ b/pypika/dialects.py
@@ -26,7 +26,7 @@ from pypika.utils import (
 
 class SnowFlakeQueryBuilder(QueryBuilder):
     QUOTE_CHAR = None
-    ALIAS_QUOTE_CHAR = '"'
+    ALIAS_QUOTE_CHAR = ''
     QUERY_ALIAS_QUOTE_CHAR = ''
 
     def __init__(self, **kwargs: Any) -> None:
--- a/pypika/tests/dialects/test_snowflake.py
+++ b/pypika/tests/dialects/test_snowflake.py
@@ -13,7 +13,7 @@ class QuoteTests(unittest.TestCase):
     def test_use_double_quotes_on_alias_but_not_on_terms(self):
         q = SnowflakeQuery.from_(self.table_abc).select(self.table_abc.a.as_("bar"))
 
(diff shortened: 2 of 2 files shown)


mikeengland:
Unfortunately I think this will cause some issues, for example, if you have an alias with non-alphanumeric characters e.g. `select firstname "$userfirstname" from users`. Without the double quotes I believe Snowflake will throw an error.

zippeurfou:
Yes that's correct a better fix is needed. I can close the PR if you want.

mikeengland:
Hi @zippeurfou, we'd need to at least support use cases where alias' are both wrapped in quotes and not wrapped. I agree about the test, feel free to add it if you have any spare time!

PR #489, opened 2020-09-24
Add DDL for temporal tables

rob-blackbourn:
The following changes add data definitions for temporal tables along with a bunch of other stuff for data definition.

## Columns

* nullable - an optional `bool` argument which (if set) will emit `NULL` or `NOT NULL`,
* default - an optional `Term` which (if set) will emit `DEFAULT <value>`

## Unique

A new builder method taking a list of Union[str, Column]; e.g.

```python
Query.create_table('t').columns(Column('c1', 'INT'),Column( 'c2', 'VARCHAR(100)')).unique('c1', 'c2').unique('c2')
```

This will emit:

```sql
CREATE TABLE t (
  c1 INT,
  c2 VARCHAR(100),

  UNIQUE (c1, c2),
  UNIQUE (c2)
)
```

## Primary Key

The primary key can be set as follows:

```python
Query.create_table('t').columns(Column('c1', 'INT'),Column( 'c2', 'VARCHAR(100)')).primary_key('c1')
```

This will emit:

```sql
CREATE TABLE t (
  c1 INT,
  c2 VARCHAR(100),

  PRIMARY KEY (c1)
)
```

## Period For

Multiple period for clauses can be added as follows.

```python
Query.create_table('t').columns(Column('valid_from', 'DATETIME'),Column( 'valid_to', DATETIME')).period_for('valid_period', 'valid_from', 'valid_to')
```

This will emit:

```sql
CREATE TABLE t (
  valid_from DATETIME,
  valid_to DATETIME,

  PERIOD FOR valid_period (valid_from, valid_to)
)
```

## With system versioning

System versioning can be added as follows.

```python
Query.create_table('t').columns(Column('c1', 'INT'),Column( 'c2', 'VARCHAR(100)')).with_system_versioning()
```

This will emit:

```sql
CREATE TABLE t (
  c1 INT,
  c2 VARCHAR(100),
) WITH SYSTEM VERSIONING
```


rob-blackbourn:
Hi,

First I wanted to check if this was the sort of pull request you are interested in?

I can see you're more focused on the query side.

Rob

mikeengland:
Hi @rob-blackbourn - thanks for the contribution, especially the detailed PR description! Given we already have basic support for creating tables, this PR is relevant. 

I did notice that we do not have any documentation for the create table operations. If you do have any spare time it would be very useful to have some documentation on the base create table, including your new additions and the Vertica create table support.

Is this PR considered complete now? If so, I'm happy to merge it in.

pypika/queries.py
```diff
@@ -283,6 +300,24 @@ def make_columns(*names: Union[TypedTuple[str, str], str]) -> List[Column]:
...
+class PeriodFor:
+    def __init__(self, name: str, start_column: Union[str, Column], end_column: Union[str, Column]) -> None:
+        self.name = name
+        self.start_column = start_column if isinstance(start_column, Column) else Column(start_column)
+        self.end_column = end_column if isinstance(end_column, Column) else Column(end_column)
+
+    def get_sql(self, **kwargs: Any) -> str:
+        quote_char = kwargs.get("quote_char")
+
+        period_for_sql = "PERIOD FOR {name} ({start_column_name},{end_column_name})".format(
```
mikeengland, line 312: Is this syntax MSSQL only? 

MariaDB in this example wraps the start and end columns in a function https://mariadb.com/kb/en/temporal-data-tables/

rob-blackbourn: Hi @mikeengland 

I wondered about this, but it also exists in oracle and mssql. Oracle works as MariaDB, but for mssql the only valid period name is `SYSTEM_TIME`: e.g.

```sql
CREATE TABLE t
(
  id INT,
  valid_from DATETIME2,
  valid_to DATETIME2,
  PERIOD FOR SYSTEM_TIME (valid_from, valid_to)
) WITH SYSTEM VERSIONING
```

mikeengland: I don't think it is an issue that needs to be addressed now, but it should be easy enough to create a custom create query builder where needed (like we do for Vertica) and override the period_for function so that it uses a PeriodFor class which generates SQL specific to that platform.

rob-blackbourn: Cool. Would you like me to make any more changes?

mikeengland: No, it looks good. I'll merge. Thanks!

rob-blackbourn: Superb!

I'll do another PR for the docs

PR #490, opened 2020-09-26
Added some documentation for create table

rob-blackbourn:
I added a small section for creating tables in the README/tutorial section.

I added docstrngs to the create table class.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1073,6 +1073,56 @@ MSSQL:
 
 You can find out what parameter style is needed for DBAPI compliant drivers here: https://www.python.org/dev/peps/pep-0249/#paramstyle or in the DB driver documentation.
 
+Creating Tables
+^^^^^^^^^^^^^^^
+
+The entry point for creating tables is ``pypika.Query.create_table``, which is used with the class ``pypika.Column``.
+As with selecting data, first the table should be specified. This can be either a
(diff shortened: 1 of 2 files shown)


README.rst
```diff
@@ -1073,6 +1073,56 @@ MSSQL:
...
+.. code-block:: python
+
+    stmt = Query \
+        .create_table("person") \
+        .columns(
+            Column("id", "INT", nullable=False),
+            Column("first_name", "VARCHAR(100)", nullable=False),
+            Column("last_name", "VARCHAR(100)", nullable=False),
+            Column("phone_number", "VARCHAR(20)", nullable=True),
+            Column("status", "VARCHAR(20)", nullable=False, default=ValueWrapper("NEW")),
```
mikeengland, line 1093: Thanks for adding these docs. 

Out of interest, if a non-wrapped data type is passed e.g. default='new' or default=1, does the code behind the scenes wrap this with a ValueWrapper, or does the current implementation require you to pass a wrapped value?
It'd be best if the wrapping occurred behind the scenes to provide a more simple API.

mikeengland (approved):

README.rst
rob-blackbourn: It doesn't wrap it at the moment. I can do this in my next request.

I'm gradually getting a better feel for how things work!

PR #491, opened 2020-09-26
Feature/temporal query support

rob-blackbourn:
I've added support for temporal table querying. I built this on top of my previous pull request BTW :(

I'm not sure how close this is to how you would like it done.

There are a bunch of examples in the README/tutorial near the bottom. The approach I've taken is to add the criteria to the `Table`. Here's an example:

```python
t = Table("abc")
q = Query.from_(t.for_(SYSTEM_TIME.as_of('2020-01-01'))).select("*")
```

This produces:

```sql
'SELECT * FROM "abc" FOR SYSTEM_TIME AS OF '2020-01-01'
```

There is support for all the valid operations including `UPDATE` and `DELETE`.

A side affect of applying the criteria to the `Table` is that joins are also supported.

Rob

### DIFF
--- a/pypika/__init__.py
+++ b/pypika/__init__.py
@@ -75,6 +75,7 @@ from pypika.terms import (
     JSON,
     Not,
     NullValue,
+    SystemTimeValue,
     Parameter,
     Rollup,
     Tuple,
@@ -95,3 +96,6 @@ from pypika.utils import (
 __author__ = "Timothy Heys"
 __email__ = "theys@kayak.com"
 __version__ = "0.39.1"
+
+NULL = NullValue()
+SYSTEM_TIME = SystemTimeValue()
--- a/pypika/__init__.py
+++ b/pypika/__init__.py
@@ -95,7 +95,7 @@ from pypika.utils import (
 
 __author__ = "Timothy Heys"
 __email__ = "theys@kayak.com"
-__version__ = "0.41.0"
+__version__ = "0.40.0"
 
 NULL = NullValue()
 SYSTEM_TIME = SystemTimeValue()
--- a/pypika/__init__.py
+++ b/pypika/__init__.py
@@ -95,7 +95,7 @@ from pypika.utils import (
 
 __author__ = "Timothy Heys"
 __email__ = "theys@kayak.com"
-__version__ = "0.39.1"
+__version__ = "0.40.0"
 
 NULL = NullValue()
 SYSTEM_TIME = SystemTimeValue()
--- a/pypika/__init__.py
+++ b/pypika/__init__.py
@@ -95,7 +95,7 @@ from pypika.utils import (
 
 __author__ = "Timothy Heys"
 __email__ = "theys@kayak.com"
-__version__ = "0.40.0"
+__version__ = "0.41.0"
 
 NULL = NullValue()
 SYSTEM_TIME = SystemTimeValue()
--- a/pypika/tests/test_selects.py
+++ b/pypika/tests/test_selects.py
@@ -334,7 +334,7 @@ class SelectTests(unittest.TestCase):
                 str(q)
             )
 
-        with self.subTest("with as of ALL"):
+        with self.subTest("with ALL"):
             q = Query.from_(t.for_(SYSTEM_TIME.all_())).select("*")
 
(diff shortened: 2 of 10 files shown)


pypika/__init__.py
```diff
@@ -95,3 +96,6 @@
 __author__ = "Timothy Heys"
 __email__ = "theys@kayak.com"
 __version__ = "0.39.1"
+
+NULL = NullValue()
+SYSTEM_TIME = SystemTimeValue()
```
mikeengland, line 101: There is a missing newline here. We still need to set up Black formatting at some point.

pypika/tests/test_selects.py
```diff
@@ -302,6 +303,73 @@ def test_table_select_alias_with_offset_and_limit(self):
...
+            q = Query.from_(
+                t.for_(SYSTEM_TIME.from_to('2020-01-01', '2020-02-01'))
+            ).select("*")
+
+            self.assertEqual(
+                'SELECT * FROM "abc" FOR SYSTEM_TIME FROM \'2020-01-01\' TO \'2020-02-01\'',
+                str(q)
+            )
+
+        with self.subTest("with as of ALL"):
```
mikeengland, line 337: Minor: Should just be "with ALL"

pypika/tests/test_selects.py
```diff
@@ -302,6 +303,73 @@ def test_table_select_alias_with_offset_and_limit(self):
...
+            q = Query.from_(
+                t.for_(t.valid_period.from_to('2020-01-01', '2020-02-01'))
+            ).select("*")
+
+            self.assertEqual(
+                'SELECT * FROM "abc" FOR "valid_period" FROM \'2020-01-01\' TO \'2020-02-01\'',
+                str(q)
+            )
+
+        with self.subTest("with period as of ALL"):
```
mikeengland, line 365: Same as above

mikeengland:
Looks like you have a merge conflict in the PR

pypika/tests/test_selects.py
```diff
@@ -302,6 +303,73 @@ def test_table_select_alias_with_offset_and_limit(self):
               Query.from_("abc").select("foo")[10:10],
         )
 
+    def test_temporal_select(self):
+        t = Table("abc")
```
mikeengland, line 307: Would you also mind adding a test for a use case with a join?

rob-blackbourn:
Hi @mikeengland 

I think I've fixed all the things you spotted.

It's worth checking the example of joins I put in the tutorial, as it describes a potential gotcha regarding joins that you might not be happy with, although I think it makes sense.

rob-blackbourn:
I'm a bit rubbish at git. How do I resolve this conflict?

mikeengland:
> I'm a bit rubbish at git. How do I resolve this conflict?

I'd use PyCharm's merge conflict feature by doing  `git pull --rebase upstream master` and then resolving the conflicts inside of PyCharm. It seems you can now also resolve the conflicts inside of Github inside of this PR, but I haven't used that before!

x8lucas8x:
> I'm a bit rubbish at git. How do I resolve this conflict?

If you pull again from master, inside your feature branch, you will see the conflicts locally. So you can just fix the conflicts and update this PR. The conflicts seem pretty basic and were caused by a previous black formatting, which your PR doesn't have applied at the time.

rob-blackbourn:
I used the GitHub GUI like a child :)

rob-blackbourn:
I've formatted with black

rob-blackbourn:
Let me know if there's anything more you need me to do with this.

pypika/__init__.py
```diff
@@ -95,7 +95,7 @@
 
 __author__ = "Timothy Heys"
 __email__ = "theys@kayak.com"
-__version__ = "0.40.0"
+__version__ = "0.41.0"
```
mikeengland, line 98: Hi @rob-blackbourn, would you mind reverting this as we use the bumpversion tool to manage the versioning, which also updates setup.cfg.

rob-blackbourn: @mikeengland done

mikeengland:
Thanks!

PR #492, opened 2020-09-28
Change docs/4_extending.srt

SuperBo:
Fix documents about extending functions.


mikeengland:
Hi @SuperBo - I am not sure what exactly this is fixing. Is there an issue with the current documentation?

Function and AnalyticFunction can both be imported from the top level as specified in the existing documentation e.g. from pypika import Function.

SuperBo:
@mikeengland : You can' t import them directly with latest version (0.42.1)

> ImportError: cannot import name 'Functions' from 'pypika' (/Users/y.nguyen/Workspace/venv/data/lib/python3.8/site-packages/pypika/__init__.py)

mikeengland:
My mistake, thanks for fixing!

PR #494, opened 2020-09-28
Inserting and updating data support key-value pairs

Zhiwei1996:
Hello, firend

`pypika` is awesome, thanks for your creation

In many cases, inserting and updating data use dict directly is convenient, and then I add features for it

What do you think?

### DIFF
--- a/pypika/__init__.py
+++ b/pypika/__init__.py
@@ -100,7 +100,7 @@ from pypika.utils import (
 
 __author__ = "Timothy Heys"
 __email__ = "theys@kayak.com"
-__version__ = "0.44.1"
+__version__ = "0.46.0"
 
 NULL = NullValue()
 SYSTEM_TIME = SystemTimeValue()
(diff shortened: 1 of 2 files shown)


stephane-klein:
@Zhiwei1996 @mikeengland some news about this Pull Request ?

gl3nn:
While I like the general idea, I'm not a big fan of the implementation. Especially the insert needs a different API.

The set_many I could be convinced by, if it reused the implementation of many.

@stephane-klein What do you need from this PR?

Zhiwei1996:
> While I like the general idea, I'm not a big fan of the implementation. Especially the insert needs a different API.
> 
> 
> 
> The set_many I could be convinced by, if it reused the implementation of many.
> 
> 
> 
> @stephane-klein What do you need from this PR?

I also think the implementation is not elegant，but the idea is useful
Can you implement it, or give some suggestions, I will reimplement

gl3nn:
Did you remove your commit? Github has incorrectly marked this as merged.

PR #496, opened 2020-09-29
Added value wrapping for column defaults

rob-blackbourn:
I think this is the change you wanted @mikeengland 

Before this column defaults had to be a `Term`, meaning that a simple value had to be wrapped. e.g.

```python
a = Column("a", "INT", default=ValueWrapper(42))
```

This change will wrap non `None` `Term`s

```python
a = Column("a", "INT", default=42)
```


mikeengland:
Thanks for fixing this!

PR #498, opened 2020-10-08
Add support for insert queries with only default values

moorecode:
Hello,

I wanted to be able to generate insert queries that made use of 'DEFAULT VALUES'; i.e `'INSERT INTO "abc" DEFAULT VALUES' `I've implemented it here. I've done my best to think of edge cases or inconsistencies but I expect there may be some corrections needed, please let me know and I'll make those changes.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -857,11 +857,9 @@ Rows can be inserted into a table with default values for all columns
 
     customers = Table('customers')
 
-    q = Query.into(customers).insert_default_values()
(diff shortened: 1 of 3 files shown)


x8lucas8x:
I'm afraid this contribution addresses a very specific use case (i.e. only when all values must be default). As is, I think it isn't worth it. Perhaps if there was a reasonable syntax to apply default only partially. That said, using LiteralValue('DEFAULT') as a constant in code wouldn't be hard either. So that one could apply DEFAULT to only certain columns in the provided tuple.

Issue #508, opened 2020-10-22
Arithmetic Order of Operations Bug


anthonyaag:
Hi All, 

When using parens in statements I would expect them to be carried forward into the generated sql to ensure that my intended order of operations was maintained. 

I'd expect that 
```
from pypika.terms import Term

example             = Term.wrap_constant(1.0) /  Term.wrap_constant(2.0) * Term.wrap_constant(3.0)
example_with_parens = Term.wrap_constant(1.0) / (Term.wrap_constant(2.0) * Term.wrap_constant(3.0))
```

would generate different sql representations `1.0/2.0*3.0` and `1.0/(2.0*3.0)` respectively. However, that does not appear to be the case. 

It seems that parentheses are being ignored when constructing sql for division and multiplication. I see this is done so explicitly in `ArithmeticExpression`.  I think this causes a correctness issue when parens are used, as the above examples produce identical sql. 

```
print((example.get_sql(),
       example_with_parns.get_sql(),
       example.get_sql() == example_with_parns.get_sql()))
> ('1.0/2.0*3.0', '1.0/2.0*3.0', True)
```

This results in both `example` and `example_with_parens` returning `1.50` when executed, but the `example_with_parens` should return  `0.166`.

gl3nn:
Fair point. Would definitely accept a PR for this.

mkbsh:
I can take this. Working on a PR now.

knowsuchagency:
Should this issue be closed now that #532 has been merged?

PR #515, opened 2020-11-03
Enclose CASE statement in parentheses

cpinamtz:
Just as a dumb idea to deal with this problem that other users have already mentioned (#508). In my case, I need this for the `CASE` statement.

### DIFF
--- a/pypika/terms.py
+++ b/pypika/terms.py
@@ -990,10 +990,11 @@ class ArithmeticExpression(Term):
 
 
 class Case(Term):
-    def __init__(self, alias: Optional[str] = None) -> None:
+    def __init__(self, alias: Optional[str] = None, parentheses=False) -> None:
         super().__init__(alias=alias)
         self._cases = []
         self._else = None
+        self.parentheses = parentheses
 
     def nodes_(self) -> Iterator[NodeT]:
         yield self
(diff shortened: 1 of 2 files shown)


gl3nn:
I don't think this relates to the linked issue (as there it seems to be related to order of operations) but if this is needed for certain platforms then we can merge it in. What platform are you targeting?

cpinamtz:
Well it's true it's not totally related but there is a problem trying to use parentheses depending of the needs of your query. 

I'm working with Oracle

cpinamtz:
For those who need to enclose your `CASE` statements when you're using them in **arithmetical expression** like the following:
```sql 
(CASE WHEN cond1 THEN then1 ELSE else1 END) + 
(CASE WHEN cond2 THEN then2 ELSE else2 END) + 
(CASE WHEN cond3 THEN then3 ELSE else3 END) + 
...
```
You can really enclose between parenthesis your `CASE` statements with the following workaround:
```python
your_sum += 1 * (0 + Case().when(cond, sta1).else_(sta22))
```
To understand why, check [the `ArithmeticExpression`'s `get_sql` method](https://github.com/kayak/pypika/blob/3c33b639350cbaa1c7cfba04aeab5ef8c590d1b0/pypika/terms.py#L980-L984) 

Anyways, I'm not going to close this issue 'cause I don't know if there is someone has a trickier situation where this workaround is not feasible. Up to you :ok_hand:

gl3nn:
I looked up some Oracle documentation and it doesn't seem like parentheses are needed at all : https://docs.oracle.com/cd/B19306_01/server.102/b14200/expressions004.htm . I believe the test you added in the PR would also work without parentheses for Oracle. Are you using the case statement in an arithmetical expression? (That wasn't clear before but would explain the link with the ticket.) If that's the case then I don't think this PR makes sense and then it should be solved on the arithmetic level.

Issue #394, opened 2020-04-08
Oracle Query Builder Non Functional


reedjosh:
I am trying to add Oracle support to <a href="https://github.com/tortoise/tortoise-orm">Tortoise ORM</a>. See PR https://github.com/tortoise/tortoise-orm/pull/341

The Oracle query builder seems to be just a pass through and is not customized for Oracle at all.

Some examples I've been working around include:
Limit functionality/Offset Functionality.
https://github.com/kayak/pypika/issues/211

Quotechars used prevent the query from working.
https://github.com/kayak/pypika/issues/37

My current work around is to build my own OracleQuery class as follows:
``` python
class OracleQueryBuilder(QueryBuilder):  # type: ignore
    def __init__(self, **kwargs):
        super(OracleQueryBuilder, self).__init__(dialect=Dialects.ORACLE, **kwargs)

    def get_sql(self, *args, **kwargs):
        return super(OracleQueryBuilder, self).get_sql(*args, groupby_alias=False, **kwargs)

    def _limit_sql(self):
        return f" FETCH NEXT {self._limit} ROWS ONLY"


class OracleQuery(Query):  # type: ignore
    """
    Defines a query class for use with Oracle.
    """

    @classmethod
    def _builder(cls, **kwargs):
        builder = OracleQueryBuilder(**kwargs)  # type: ignore
        builder.QUOTE_CHAR = ""
        return builder
```

Could some expanded form of this be integrated into PyPika itself?

twheys:
Hi, please take a look at this issue about limits in oracle and let me know if it's a acceptable approach for you https://github.com/kayak/pypika/issues/211

Generally we try to avoid overriding functions for different dialects to yield different results in favor of yielding a python API that matches the resulting SQL. If you prefer to use the syntax `FETCH NEXT {self._limit} ROWS ONLY`, could you please open a PR adding some function with a similar name to the OracleQuery class?

reedjosh:
#211 is the issue I linked initially. My main goal is having a consistent function to call across the dialects. Right now calling limit on the OracleDialect breaks the resulting query.
 
So if the limit function added `WHERE ROWNUM <= 2`, that would be fine, but I'm not sure how to implement that so that it gets added properly.

Mostly I was hoping since this wouldn't be in the core of Pypika we would be able to integrate the workaround as above.

Otherwise the only way to use Pypika is to workaround the limit function downstream.

twheys:
[...]

Issue #298, opened 2019-08-16
get_sql() throws exception on MSSQLQuery that includes where clause with subquery


paul-ko:
I've noticed that when I use the `MSSQLQuery` dialect to construct a query including a where predicate such as `where SomeColumn not in (select SomeOtherColumn from SomeTable)`, `get_sql()` throws what appears to be a spurious exception.  If I swap in a `Query` instead of the `MSSQLQuery`, no such exception occurs, and the generated query runs without issue against a SQL Server database.

Version: 0.33.1, installed via pip
Stack frames from within pypika:

```
  File "<snip>\venv\lib\site-packages\pypika\queries.py", line 836, in __str__
    return self.get_sql(quote_char=self.quote_char, dialect=self.dialect)
  File "<snip>\venv\lib\site-packages\pypika\dialects.py", line 308, in get_sql
    return super(MSSQLQueryBuilder, self).get_sql(*args, groupby_alias=False, **kwargs)
  File "<snip>\venv\lib\site-packages\pypika\queries.py", line 936, in get_sql
    querystring += self._where_sql(**kwargs)
  File "<snip>\venv\lib\site-packages\pypika\queries.py", line 1034, in _where_sql
    return ' WHERE {where}'.format(where=self._wheres.get_sql(quote_char=quote_char, subquery=True, **kwargs))
  File "<snip>\venv\lib\site-packages\pypika\terms.py", line 708, in get_sql
    container=self.container.get_sql(**kwargs),
  File "<snip>\venv\lib\site-packages\pypika\dialects.py", line 308, in get_sql
    return super(MSSQLQueryBuilder, self).get_sql(*args, groupby_alias=False, **kwargs)
TypeError: get_sql() got multiple values for keyword argument 'groupby_alias'
```

Ugly demonstration code:

```
from pypika import Field, MSSQLQuery, Query, Table


def generate_query(query_class):
    tbl = Table('query_tbl', schema='scm')
    col = Field('query_col', table=tbl)
    tbl2 = Table('sub_tbl', schema='scm')
    col2 = Field('sub_col', table=tbl2)

    print(query_class.__name__)
    subquery = query_class().from_(tbl2).select(col2)
    print('Subquery:', subquery)
    main_query_base = query_class().from_(tbl).select(col)
    print('Main query base:', main_query_base)
    predicate = col.notin(subquery)
    print('Predicate:', predicate)
    full_query = main_query_base.where(predicate)
    try:
        print('Fullquery:', full_query)
[...]

PR #516, opened 2020-11-04
Fix OracleQuery Dialect Quote Char

noambloom:
Oracle SQL query breaks when double quoted

A quoted identifier begins and ends with double quotation marks (").
If you name a schema object using a quoted identifier,
then you must use the double quotation marks whenever you refer to that object.
Thus it makes sense that the user will double quote the identifier
him/herself.

A nonquoted identifier is not surrounded by any punctuation.

See https://docs.oracle.com/database/121/SQLRF/sql_elements008.htm

Addresses https://github.com/kayak/pypika/issues/394#issuecomment-616012215
Addresses #298 for OracleQuery

### DIFF
--- a/pypika/dialects.py
+++ b/pypika/dialects.py
@@ -270,6 +270,8 @@ class VerticaQuery(Query):
 
 
 class OracleQueryBuilder(QueryBuilder):
+    QUOTE_CHAR = None
+
     def __init__(self, **kwargs: Any) -> None:
         super().__init__(dialect=Dialects.ORACLE, **kwargs)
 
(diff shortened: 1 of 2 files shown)


pypika/dialects.py
```diff
@@ -274,7 +274,9 @@ def __init__(self, **kwargs: Any) -> None:
         super().__init__(dialect=Dialects.ORACLE, **kwargs)
 
     def get_sql(self, *args: Any, **kwargs: Any) -> str:
-        return super().get_sql(*args, groupby_alias=False, **kwargs)
+        kwargs.setdefault("quote_char", None)
```
gl3nn, line 277: It's probably better to use `QUOTE_CHAR = None` like how `class SnowFlakeQueryBuilder` does it. Could you double check whether that results in the same desired behaviour?

noambloom: Works :)

gl3nn (approved):

Issue #464, opened 2020-07-09
Updating records from values


NirAlt:
Hi,
I'm trying to create something like the following with PyPika:

update test as t set
    column_a = c.column_a,
    column_c = c.column_c
from (values
    ('123', 1, '---'),
    ('345', 2, '+++')  
) as c(column_b, column_a, column_c) 
where c.column_b = t.column_b;

The idea is to update specific fields of multiple records from values with a clause on one of the columns to match a "column" in the values.

Is this possible?

Thanks!

PR #517, opened 2020-11-06
Add Values class for '(VALUES ...) as alias(field_name, ...)' expressions

danifus:
Adds a preliminary implementation of the feature in #464

The PR isn't finished yet but I'm looking for some feedback as to whether this is something you would like to incorporate into the project before I give it a bit of polish. I wrote it with MSSQL in mind which is a bit more restrictive than other databases so I'll have to generalize it.

Also let me know if there is a better way of approaching this and I can rework the patch to use the better approach.

Some thoughts:
- Is it worthwhile splitting the column alias functionality into another feature? Postgres allows renaming table columns in a `from` alias

Some documentation for easy reference:
Compatibility and support: https://modern-sql.com/feature/values
Column aliases: https://modern-sql.com/feature/table-column-aliases

- postgres:
  https://www.postgresql.org/docs/9.5/sql-values.html

- mssql:
  Column alias is required
  https://docs.microsoft.com/en-us/sql/t-sql/queries/table-value-constructor-transact-sql?view=sql-server-ver15

- mysql:
  Their 'row constructor list' requires the `ROW` key word. Mariadb doesn't allow the row keyword
   https://dev.mysql.com/doc/refman/8.0/en/values.html
   https://mariadb.com/kb/en/table-value-constructors/

- sqlite:
  Some info spread across multiple pages. In short, it is supported but column renaming is only allowed in a 'with' clause.


AzisK:
Could you update the code with the newest master branch code so that the CI tests kick in?

PR #531, opened 2020-12-09
Order nulls last options

ferrants:
Is there a better way to add this support? I was looking for an `Order.desc.nulls_last` originally

### DIFF
--- a/pypika/enums.py
+++ b/pypika/enums.py
@@ -46,6 +46,8 @@ class Boolean(Comparator):
 class Order(Enum):
     asc = "ASC"
     desc = "DESC"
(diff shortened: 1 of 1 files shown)


pypika/enums.py
```diff
@@ -46,6 +46,8 @@ class Boolean(Comparator):
 class Order(Enum):
     asc = "ASC"
     desc = "DESC"
+    asc_nulls_last = "ASC NULLS LAST"
```
ferrants, line 49: This isn't supported in all DB types, it works in Postgres though, not sure if this should even be here

ferrants:
Not going to use this framework. Closing.

PR #536, opened 2021-01-04
Fix PostgreSQLQueryBuilder returning clause code causing Py3.9 test failures

mikeengland:
This PR fixes the Python 3.9 test failures in the Github Actions CI. 

This was previously passing due to this statement: `if field.table not in {self._insert_table, self._update_table} and term not in self._from`, specifically the second statement. In my Python 3.6 environment, the Term `__eq__` method was called, always returning a truthy value (as this returns a BasicCriterion object, intended for query building). In Python 3.9, this is not called (I couldn't find anything in the release notes though). Therefore, I have rewritten this to check if the field's table is one of the base and join tables.

FYI @abondar

### DIFF
--- a/pypika/dialects.py
+++ b/pypika/dialects.py
@@ -1,3 +1,4 @@
+import itertools
 from copy import copy
 from typing import Any, Optional, Union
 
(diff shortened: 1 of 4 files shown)


abondar:
Cool, thank you! 
Will try returning 3.9 to tortoise test matrix once this is released

gl3nn (approved):

PR #538, opened 2021-01-07
Add the glob matching comparator (used in SQLite)

kapily:
SQLite has a `GLOB` operator. It is a case-sensitive version of the `LIKE` operator. 

Documentation here:
https://sqlite.org/lang_expr.html
https://stackoverflow.com/questions/37508793/whats-the-difference-between-like-and-glob-in-sqlite

relevant quote:
> The GLOB operator is similar to LIKE but uses the Unix file globbing syntax for its wildcards. Also, GLOB is case sensitive, unlike LIKE.

### DIFF
--- a/pypika/enums.py
+++ b/pypika/enums.py
@@ -33,6 +33,7 @@ class Matching(Comparator):
     regex = " REGEX "
     bin_regex = " REGEX BINARY "
     as_of = " AS OF "
+    glob = " GLOB "
 
 
 class Boolean(Comparator):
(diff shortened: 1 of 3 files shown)


mikeengland:
Would you mind adding a test case where a query uses this new operator? Thanks!

kapily:
@mikeengland - done! Wow super impressed by your test coverage and test code quality

mikeengland (approved):

mikeengland:
Thanks for your contribution!

mikeengland:
Released in 0.47.1

PR #539, opened 2021-01-08
Fix empty Postgres array syntax

### LINKED ISSUE #527: query.get_sql(dialect=Dialects.POSTGRESQL) does not provide correct *empty* arrays for postgresql
When passing it an array to insert such as `[]` it will convert to `ARRAY[]` which returns the following error in PostgreSQL:
```
ERROR:  cannot determine type of empty array at character XXX
HINT:  Explicitly cast to the desired type, for example ARRAY[]::integer[].
```

The correct thing to do is to return `'{}'` instead of casting to a specific type that is probably unknown to Pypika. `'{}'` gives PSQL the right to cast it according to the column type.

Apexal:
Closes #527 

Empty arrays for Postgres and Redshift are switched from `ARRAY[]` to `'{}'` so that PSQL can cast it to the right column type. This would previously give the following error: 
```
ERROR:  cannot determine type of empty array at character XXX
HINT:  Explicitly cast to the desired type, for example ARRAY[]::integer[].
```

### DIFF
--- a/pypika/terms.py
+++ b/pypika/terms.py
@@ -611,7 +611,10 @@ class Array(Tuple):
     def get_sql(self, **kwargs: Any) -> str:
         dialect = kwargs.get("dialect", None)
         values = ",".join(term.get_sql(**kwargs) for term in self.values)
-        sql = ("ARRAY[{}]" if dialect in (Dialects.POSTGRESQL, Dialects.REDSHIFT) else "[{}]").format(values)
+        if dialect in (Dialects.POSTGRESQL, Dialects.REDSHIFT):
+            sql = "ARRAY[{}]".format(values) if len(values) > 0 else "'{}'"
+        else:
+            sql = "[{}]".format(values)
         return format_alias_sql(sql, self.alias, **kwargs)
 
 
--- a/pypika/terms.py
+++ b/pypika/terms.py
@@ -611,10 +611,11 @@ class Array(Tuple):
     def get_sql(self, **kwargs: Any) -> str:
         dialect = kwargs.get("dialect", None)
         values = ",".join(term.get_sql(**kwargs) for term in self.values)
+
+        sql = "[{}]".format(values)
         if dialect in (Dialects.POSTGRESQL, Dialects.REDSHIFT):
             sql = "ARRAY[{}]".format(values) if len(values) > 0 else "'{}'"
-        else:
-            sql = "[{}]".format(values)
+
         return format_alias_sql(sql, self.alias, **kwargs)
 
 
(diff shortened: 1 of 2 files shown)


pypika/terms.py
```diff
@@ -611,7 +611,10 @@ class Array(Tuple):
     def get_sql(self, **kwargs: Any) -> str:
         dialect = kwargs.get("dialect", None)
         values = ",".join(term.get_sql(**kwargs) for term in self.values)
-        sql = ("ARRAY[{}]" if dialect in (Dialects.POSTGRESQL, Dialects.REDSHIFT) else "[{}]").format(values)
+        if dialect in (Dialects.POSTGRESQL, Dialects.REDSHIFT):
+            sql = "ARRAY[{}]".format(values) if len(values) > 0 else "'{}'"
+        else:
+            sql = "[{}]".format(values)
```
airtonzanon, line 617: What do you think of declaring the default SQL as `sql = "[{}]".format(values)` and only if we have PostgreSQL or Redshift we change it.

like this:
```python
        sql = "[{}]".format(values)
        if dialect in (Dialects.POSTGRESQL, Dialects.REDSHIFT):
            sql = "ARRAY[{}]".format(values) if len(values) > 0 else "'{}'"
        return format_alias_sql(sql, self.alias, **kwargs)
```

gl3nn:
Please add a test for the specific case you are trying to solve (i.e. the case that throws an error without the fix).

pypika/terms.py
gl3nn: Is this a very specific thing where `{}` is considered an empty untyped array for postgres? As mentioned before, a test (or multiple?) could provide some clarity (and an opportunity to describe the behaviour).

Apexal: Based off of https://www.postgresql.org/docs/current/arrays.html, it seems that `'{}'` is a literal array constant just like `'{1, 2, 3}'` versus using the `ARRAY` constructor syntax `ARRAY[20000, 25000, 25000, 25000]`. Using the literal syntax seems to be preferred, as the ARRAY constructor is only mentioned offhand at the end of the section.

I tried switching the Array.get_sql to use the `'{val1, val2, val3}'` syntax but for nested arrays it generates extra single quotes: `'{'{1,2,3}','{4,5,6}'}'`. I'm guessing this is why the ARRAY constructor was used in the first place. It might make most sense to do what I originally had, with @airtonzanon 's change as it properly handles nested arrays, even though the PSQL documentation doesn't have `ARRAY` used in each inner array. I added two new array tests to check these cases. I ran the queries on a PSQL instance to check.

gl3nn (approved):

PR #541, opened 2021-01-13
Remove quotes from Snowflake CREATE/DROP queries

jasonspeck:
For consistency with SELECT queries generated by the Snowflake dialect.

### DIFF
--- a/pypika/dialects.py
+++ b/pypika/dialects.py
@@ -8,6 +8,7 @@ from pypika.queries import (
     Query,
     QueryBuilder,
     Table,
+    DropQueryBuilder,
 )
 from pypika.terms import ArithmeticExpression, Criterion, EmptyCriterion, Field, Function, Star, Term, ValueWrapper
 from pypika.utils import (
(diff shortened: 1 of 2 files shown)


pypika/dialects.py
```diff
@@ -36,6 +45,22 @@ def __init__(self, **kwargs: Any) -> None:
         super().__init__(dialect=Dialects.SNOWFLAKE, **kwargs)
 
 
+class SnowflakeCreateQueryBuilder(CreateQueryBuilder):
+    QUOTE_CHAR = None
+    QUERY_CLS = SnowflakeQuery
+
+    def __init__(self) -> None:
```
mikeengland, line 52: CreateQueryBuilder specifies ALIAS_QUOTE_CHAR=None but SnowflakeQueryBuilder specifies '"'. It is possible this could cause issues in cases where aliases are used in delete/update queries. Would you mind looking at adding the alias quote char to make it consistent with SnowflakeQueryBuilder?

pypika/dialects.py
```diff
@@ -36,6 +45,22 @@ def __init__(self, **kwargs: Any) -> None:
         super().__init__(dialect=Dialects.SNOWFLAKE, **kwargs)
 
 
+class SnowflakeCreateQueryBuilder(CreateQueryBuilder):
+    QUOTE_CHAR = None
```
mikeengland, line 49: I noticed CreateQueryBuilder specifies ALIAS_QUOTE_CHAR=None but SnowflakeQueryBuilder specifies '"'. I don't think this will cause any issues but noting this here anyway.

mikeengland (approved):

pypika/dialects.py
```diff
@@ -25,6 +26,14 @@ class SnowflakeQuery(Query):
     def _builder(cls, **kwargs: Any) -> "SnowflakeQueryBuilder":
         return SnowflakeQueryBuilder(**kwargs)
 
+    @classmethod
```
mikeengland, line 29: Any reason these are class methods over static methods?

mikeengland:
Released as pypika 0.47.3

PR #549, opened 2021-02-04
Surround subquery in the args aggregate function

magiskboy:


x8lucas8x:
That seems to be a duplicate of https://github.com/kayak/pypika/pull/566. I'd merge this if the conflicts are addressed. Or https://github.com/kayak/pypika/pull/566 in case the author adds tests.

magiskboy:
Why? This pr is more early than #566

x8lucas8x:
@magiskboy like I said there is a conflict in this, so I cannot merge it now. Either you fix it or #566 adds tests. I'll merge whatever happens first.

x8lucas8x (approved):

pypika/terms.py
```diff
@@ -1259,7 +1259,10 @@ def get_function_sql(self, **kwargs: Any) -> str:
 
         return "{name}({args}{special})".format(
             name=self.name,
-            args=",".join(self.get_arg_sql(arg, **kwargs) for arg in self.args),
+            args=",".join(
+                p.get_sql(with_alias=False, subquery=True, **kwargs) if hasattr(p, "get_sql") else str(p)
```
x8lucas8x, line 1263: Can you please keep using self.get_arg_sql(arg, **kwargs) for the non subquery case as before? Its implementation isn't exactly equal to str(p).

x8lucas8x (changes requested):
See comments.

pypika/terms.py
magiskboy: I think that this function `get_function_sql` is called if once this is a sub-query, but not all of the terms

magiskboy: because function is only used in subquery

x8lucas8x: Not sure I understood you. I mean changing:

`args=",".join(
                p.get_sql(with_alias=False, subquery=True, **kwargs) if hasattr(p, "get_sql") else str(p)` 
                
to:

`args=",".join(
                p.get_sql(with_alias=False, subquery=True, **kwargs) if hasattr(p, "get_sql") else self.get_arg_sql(arg, **kwargs)`

magiskboy: what is `arg`? I can't see it

x8lucas8x: You've changed arg to p, as you can see in diff. So that's essentially the element from the self.args, which we are iterating.

x8lucas8x: Arg is a better variable name in this case btw.

x8lucas8x (approved):
Wonderful.

PR #561, opened 2021-03-02
Properly quote enum values

### LINKED ISSUE #560: Enums not quoted
# Bug report

## Description

When using a member of `class(Enum)` or `class(str, Enum)` in a query, I expect the underlying `.value` of the member to be used when constructing a query so that the result is the same as if I had passed the literal value explicitly.

## Example

```python
from pypika import Query, Table 
from enum import Enum

class Status(Enum):
    ON = 'on'
    OFF = 'off'

t = Table('tab')
q = Query.from_(t).select(t.col1).where(t.status == Status.ON)
```

## Actual behavior

```
>>> q.get_sql()
SELECT "col1" FROM "tab" WHERE "status"=on
```

## Expected behavior

```
>>> q.get_sql()
SELECT "col1" FROM "tab" WHERE "status"='on'
```

## Version info

Python 3.7.9
PyPika 0.47.7

cbows:
Properly handle quoting of enum values by unpacking them first

Closes #560

### DIFF
--- a/pypika/enums.py
+++ b/pypika/enums.py
@@ -70,7 +70,7 @@ class SetOperation(Enum):
     minus = "MINUS"
 
 
-class DatePart(Enum):
+class DatePart(str, Enum):
     year = "YEAR"
     quarter = "QUARTER"
     month = "MONTH"
--- a/pypika/terms.py
+++ b/pypika/terms.py
@@ -336,24 +336,27 @@ class ValueWrapper(Term):
         self.value = value
 
     def get_value_sql(self, **kwargs: Any) -> str:
+        return self.get_formatted_value(self.value, **kwargs)
+
+    @classmethod
+    def get_formatted_value(cls, value: Any, **kwargs):
         quote_char = kwargs.get("secondary_quote_char") or ""
 
         # FIXME escape values
-        if isinstance(self.value, Term):
-            return self.value.get_sql(**kwargs)
-        if isinstance(self.value, Enum):
-            return self.value.value
-        if isinstance(self.value, date):
-            value = self.value.isoformat()
+        if isinstance(value, Term):
+            return value.get_sql(**kwargs)
+        if isinstance(value, Enum):
+            return cls.get_formatted_value(value.value, **kwargs)
+        if isinstance(value, date):
+            return cls.get_formatted_value(value.isoformat(), **kwargs)
+        if isinstance(value, str):
+            value = value.replace(quote_char, quote_char * 2)
             return format_quotes(value, quote_char)
-        if isinstance(self.value, str):
-            value = self.value.replace(quote_char, quote_char * 2)
-            return format_quotes(value, quote_char)
-        if isinstance(self.value, bool):
-            return str.lower(str(self.value))
-        if self.value is None:
+        if isinstance(value, bool):
+            return str.lower(str(value))
+        if value is None:
             return "null"
-        return str(self.value)
+        return str(value)
 
     def get_sql(self, quote_char: Optional[str] = None, secondary_quote_char: str = "'", **kwargs: Any) -> str:
         sql = self.get_value_sql(quote_char=quote_char, secondary_quote_char=secondary_quote_char, **kwargs)
--- a/pypika/tests/test_selects.py
+++ b/pypika/tests/test_selects.py
@@ -1,4 +1,6 @@
 import unittest
+from datetime import date
+from enum import Enum
 
 from pypika import (
     AliasedQuery,
@@ -339,8 +341,29 @@ class SelectTests(unittest.TestCase):
             self.assertEqual('SELECT * FROM "abc" FOR "valid_period" ALL', str(q))
 
 
+class MyEnum(Enum):
+    STR = "foo"
+    INT = 0
+    BOOL = True
+    DATE = date(2020, 2, 2)
+    NONE = None
+
+
 class WhereTests(unittest.TestCase):
     t = Table("abc")
+
+    def test_where_enum(self):
+        q1 = Query.from_(self.t).select("*").where(self.t.foo == MyEnum.STR)
+        q2 = Query.from_(self.t).select("*").where(self.t.foo == MyEnum.INT)
+        q3 = Query.from_(self.t).select("*").where(self.t.foo == MyEnum.BOOL)
+        q4 = Query.from_(self.t).select("*").where(self.t.foo == MyEnum.DATE)
+        q5 = Query.from_(self.t).select("*").where(self.t.foo == MyEnum.NONE)
+
+        self.assertEqual('SELECT * FROM "abc" WHERE "foo"=\'foo\'', str(q1))
+        self.assertEqual('SELECT * FROM "abc" WHERE "foo"=0', str(q2))
+        self.assertEqual('SELECT * FROM "abc" WHERE "foo"=true', str(q3))
+        self.assertEqual('SELECT * FROM "abc" WHERE "foo"=\'2020-02-02\'', str(q4))
+        self.assertEqual('SELECT * FROM "abc" WHERE "foo"=null', str(q5))
 
     def test_where_field_equals(self):
         q1 = Query.from_(self.t).select("*").where(self.t.foo == self.t.bar)
--- a/pypika/functions.py
+++ b/pypika/functions.py
@@ -6,6 +6,7 @@ from pypika.terms import (
     AggregateFunction,
     Function,
     Star,
+    LiteralValue,
 )
 from pypika.utils import builder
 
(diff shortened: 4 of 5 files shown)


cbows:
Due to failing tests I noticed another potential bug, which I fixed in  https://github.com/kayak/pypika/pull/561/commits/d099a5b6297db52b1942fc41a528e506853f06f2

mikeengland:
Please could you add some tests to cover these changes?

pypika/terms.py
```diff
@@ -342,7 +342,8 @@ def get_value_sql(self, **kwargs: Any) -> str:
         if isinstance(self.value, Term):
             return self.value.get_sql(**kwargs)
         if isinstance(self.value, Enum):
-            return self.value.value
+            self.value = self.value.value
```
mikeengland, line 345: I think it would be best if the value was reset here. Maybe this method be refactored so you can just pass in the new value, or even just instantiate another ValueWrapper with the value? e.g.
```python
return ValueWrapper(self.value.value, self.alias).get_value_sql(**kwargs)
```

What do you think>

x8lucas8x: I agree with @mikeengland. Also that introduces a side effect to get_sql, which imo could be really tricky to debug, given nobody would expect get_sql to be actually modifying the object instance.

x8lucas8x (changes requested):
See comments on pypika/terms.py.

pypika/terms.py
cbows: @mikeengland I agree. Initially I wanted to pass it as an argument to `get_value_sql` but that's not possible without refactoring like you said.

I have updated the PR with tests and some refactoring of `get_value_sql`. The idea is to be able to pass a `_value` argument when you want to call `get_value_sql` recursively. The idea is that `get_value_sql` should include the logic for quoting base types like `int`, `bool`, `str` and such. For other, more complex types (`date`, `Enum`, etc.) it handles the conversion one of the base types and then calls itself with the converted type. As an example I have refactored the `date` conversion the same way `Enum`s are handled, which basically gets rid of the duplicate logic for `format_quotes` which was already handle for `str`.

If you don't like it, I could also implement your suggestion of instantiating another `ValueWrapper`

cbows: There is only one 'hack' if you want to call it that, and that is the introduction of an `UNSET` type (which is just an `object`), so that `None` is handled properly.

pypika/terms.py
```diff
@@ -335,25 +337,27 @@ def __init__(self, value: Any, alias: Optional[str] = None) -> None:
         super().__init__(alias)
         self.value = value
```
mikeengland, line 339: I see what you are trying to do with the UNSET object to get around the valid None value, however I think extracting the contents of the `get_value_sql` into a different method would be a more simple approach.

For example, you could have a static `get_formatted_value` (or another name) method which implements the if branches that format and return the value. Then `get_value_sql` would just need to call that with the correct value: for Enum value.value or just self.value otherwise.

That would prevent the need to use the UNSET hack. Does this make sense?

pypika/tests/test_selects.py
```diff
@@ -339,9 +341,30 @@ def test_temporal_select(self):
...
+    INT = 0
+    BOOL = True
+    DATE = date(2020, 2, 2)
+    NONE = None
+
+
 class WhereTests(unittest.TestCase):
     t = Table("abc")
 
+    def test_where_enum(self):
```
mikeengland, line 355: These tests look good thanks for adding them!

pypika/terms.py
```diff
@@ -335,25 +337,27 @@ def __init__(self, value: Any, alias: Optional[str] = None) -> None:
...
 
-    def get_value_sql(self, **kwargs: Any) -> str:
+    def get_value_sql(self, _value: Any = UNSET, **kwargs: Any) -> str:
         quote_char = kwargs.get("secondary_quote_char") or ""
 
+        # allow `get_value_sql` to be called recursively
+        value = self.value if _value is UNSET else _value
```
x8lucas8x, line 344: That's honestly very hacky, to not say poor OOP design, and definitely adds a lot of complexity to this method. Adding a param in this way, to be overriding the default behaviour of this method is something I find really hard to justify.

pypika/terms.py
cbows: I rather dislike static methods, but as I understand a classmethod would do just as well and even works with inheritance and stuff?

cbows: PR is updated.

btw, how do you prefer PR updates. Can I use force-pushs directly or use fixup commits and rebase in the end?

pypika/terms.py
cbows: PR already updated. I agree this solution is a little hacky

pypika/terms.py
mikeengland: Either way. I generally like to merge in single commit so the history remains clean.

Understood regarding classmethod in this use case. Now I look at it again it makes more sense than the staticmethod approach I originally suggested given it avoids the need to handle enums differently in `get_value_sql`!

pypika/terms.py
```diff
@@ -336,24 +336,27 @@ def __init__(self, value: Any, alias: Optional[str] = None) -> None:
...
-            value = self.value.isoformat()
-            return format_quotes(value, quote_char)
-        if isinstance(self.value, str):
-            value = self.value.replace(quote_char, quote_char * 2)
+        if isinstance(value, Term):
+            return value.get_sql(**kwargs)
+        if isinstance(value, Enum):
+            return cls.get_formatted_value(value.value, **kwargs)
+        if isinstance(value, date):
+            return cls.get_formatted_value(value.isoformat(), **kwargs)
```
mikeengland, line 351: Is there any need to call this method again here?  From what I can see it doesn't offer an improvement on the old date handling. Additionally, as `get_formatted_value` is called again, it causes value.replace to be called when it wasn't before.

cbows: The idea was to convert `date` to one of the base types (`str` in this case) and then call whatever logic is implement for `str`. I think this is generally more DRY even if just slightly. So if you change the way you handle strings, you don't need to change it again for `date`, because `date` reuses the same logic.

mikeengland (approved):

pypika/terms.py
cbows: It might be overkill here, but I thought maybe other types are added later (just like Enum was now) and then we just apply the same logic.

pypika/enums.py
```diff
@@ -70,7 +70,7 @@ class SetOperation(Enum):
     minus = "MINUS"
 
 
-class DatePart(Enum):
+class DatePart(str, Enum):
```
x8lucas8x, line 73: Why have you made this inherit from str?

cbows: Good catch. I thought there was a reason for it initially, but not anymore.

PR updated.

cbows: Oh sorry no. The tests are failing without. Let me see.

cbows: So, it is the reason I initially stated in the commit message. Here is an excerpt:

> In order to fix it, `date_part` is wrapped into a `LiteralValue` to prevent
> quoting. To allow the `DatePart` enum to work also, it is derived from `str` as
> well as `Enum`, thus inheriting the required properties.

Here is a little more explanation:
Before Enum values were quoted properly, the `Extract` function was working fine, but only because `DatePart.year` for example wasn't quoted and stayed `YEAR` instead of `'YEAR'`. The same thing happened if you had called `Extract` with a string literal, instead of one of the Enum members (like `Extract("YEAR")`). Calling it like that produced invalid SQL but wasn't tested, instead relied on the old Enum behavior. To get around the issue, I wrap `date_part` in `LiteralValue` which only accepts strings, so it would fail with `DatePart.year`. Now strings worked, but the old behavior using the Enum doesn't. Here comes the `class DatePart(str, Enum)` making it work with enum members (old behavior) and strings (new behavior).

x8lucas8x:
@cbows hey, I've replied you on https://github.com/kayak/pypika/issues/560. Although we value your PR, we now think there are better ways to achieve what you want. Please check the GitHub issue out and feel free to ask more questions. For now I'm closing this.

cbows:
@x8lucas8x I have removed the changes regarding `DatePart` from this PR and described the problem in more detail in a separate issue (https://github.com/kayak/pypika/issues/574) along with another PR that solves the issue without adding unnecessary complexity (https://github.com/kayak/pypika/pull/575). When #574 gets fixed (either by merging #575 or by providing a different solution) there is nothing blocking this PR anymore.

x8lucas8x:
@cbows #575 was merged but I cannot reopen this PR. I guess because you force push to it while it was closed. So please re-create this.

PR #563, opened 2021-03-02
Adding pypika/pivot.py with pivot function

adamddav:
Wasn't really sure where this should go so I made a new file.  The behavior is similar to this PIVOT method, the difference is that PIVOT implicitly groups on all other columns, whereas with this pivot there is no implicit grouping.

https://docs.snowflake.com/en/sql-reference/constructs/pivot.html

### DIFF
--- a/pypika/pivot.py
+++ b/pypika/pivot.py
@@ -1,0 +1,16 @@
+from typing import List, Tuple, Type
+
+from pypika import Case, Field, NULL
+import pypika.functions as fn
+
+
+def pivot(
+    pivot_column: Field,
+    value_column: Field,
+    pivot_values: List[str],
+    aggregate_function: Type[fn.AggregateFunction],
+) -> Tuple[fn.AggregateFunction]:
+    return (
+        aggregate_function(Case().when(pivot_column == pivot_value, value_column).else_(NULL)).as_(pivot_value)
+        for pivot_value in pivot_values
+    )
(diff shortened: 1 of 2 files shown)


adamddav:
is there an easy way to squash these 4 commits down into one via GitHub? or do I have to do that locally and force push?

mikeengland:
Hi @adamddav, as this is not a python representation of a built-in database function I don't think it makes sense to store this in Pypika. This feels more like a useful utility to be stored in your own code base. Therefore I will close this PR. Thanks!

adamddav:
@mikeengland, just to clarify, you are saying that it is not a representation of a built-in database function because, unlike the built-in database function, it does not perform an implicit grouping on the other columns in the table?  If it did perform this implicit grouping, and therefore faithfully replicate the built-in database function, would you still not want it in the Pypika codebase?

mikeengland:
In Pypika we are aiming to create a pythonic API for sql generation. So this is why you see function classes that are named the same as the real sql functions and an API which mimics how SQL is built e.g. `.select`, `.where` etc.

If you created a Pivot function class e.g. in pypika.functions which built a Pivot function as shown here: https://docs.snowflake.com/en/sql-reference/constructs/pivot.html, that would be acceptable as it matches the Snowflake API. However, utilities for other more custom use cases should be stored in other packages, or we would end up with a lot of custom utilities added to Pypika.

adamddav:
Understood @mikeengland , thank you.  I will not pursue this at the moment, since this utility is perfect for my usecase, and I agree, should remain in my own codebase.

If any issues every come up requesting Pivot to be added, feel free to ping me and I will refactor the logic into a `pypika.functions.Pivot` class!

PR #565, opened 2021-03-05
add nowait, skip_locked, of to for_update

long2ice:


long2ice:
So what do I need do more?

pypika/queries.py
```diff
@@ -1329,9 +1336,15 @@ def _distinct_sql(self, **kwargs: Any) -> str:
 
         return distinct
 
-    def _for_update_sql(self) -> str:
+    def _for_update_sql(self, **kwargs) -> str:
         if self._for_update:
             for_update = ' FOR UPDATE'
+            if self._for_update_of:
+                for_update += f' OF {", ".join([Table(item).get_sql(**kwargs) for item in self._for_update_of])}'
```
x8lucas8x, line 1343: Is FOR UPDATE OF ansi SQL? I couldn't find support on Mysql for it, but I could be wrong. NOWAIT and SKIP LOCKED looks good in terms of support (i.e. mysql, postgres, oracle).

pypika/tests/test_selects.py
```diff
@@ -355,6 +355,18 @@ def test_where_field_equals_for_update(self):
...
 
+    def test_where_field_equals_for_update_nowait(self):
+        q = Query.from_(self.t).select("*").where(self.t.foo == self.t.bar).for_update(nowait=True)
+        self.assertEqual('SELECT * FROM "abc" WHERE "foo"="bar" FOR UPDATE NOWAIT', str(q))
+
+    def test_where_field_equals_for_update_skip_locked(self):
+        q = Query.from_(self.t).select("*").where(self.t.foo == self.t.bar).for_update(skip_locked=True)
+        self.assertEqual('SELECT * FROM "abc" WHERE "foo"="bar" FOR UPDATE SKIP LOCKED', str(q))
+
+    def test_where_field_equals_for_update_of(self):
```
x8lucas8x, line 366: Can you please add a test with multi of values?

x8lucas8x:
This looks good. I'm just unsure whether FOR UPDATE OF is ansi SQL. If it isn't, that should be on a specific dialect instead. See https://github.com/kayak/pypika/blob/master/pypika/dialects.py if that's the case. I'll be waiting for proof that FOR UPDATE OF is ansi SQL to give an approval.

pypika/queries.py
long2ice: According to the MySQL official document, `FOR UPDATE OF` support in `8.0` but not `5.7`, see https://dev.mysql.com/doc/refman/5.7/en/select.html and https://dev.mysql.com/doc/refman/8.0/en/select.html, so I'm not sure whether shoud move to dialects

pypika/tests/test_selects.py
long2ice: OK, I will do that

x8lucas8x: Oh, sorry for not being clear enough. I meant a tuple for of kwarg with more than one value. That said, please keep the newly added tests too. It's good to have them nevertheless.

pypika/queries.py
long2ice: Maybe user should decide that themselves? We just keep now.

pypika/tests/test_selects.py
long2ice: Sorry my English is not so good, could you give a example?

x8lucas8x: I mean to add a test in which "of" kwarg has a tuple with more than one value also. Like in .for_update(of=("abc","def")).

long2ice: OK

pypika/queries.py
x8lucas8x: When I ask if FOR UPDATE OF is ANSI SQL I mean if it is part of the SQL standard. That usually implies that this syntax isn't only supported in MYSQL. Because when that's the case, it should be added to our library as part of Mysql's dialect, so that it doesn't pollute other DBs queryset with unsupported syntax. In that sense, it would be helpful if you could show proof that FOR UPDATE OF is either ANSI SQL or that well-known databases documentations (e.g. mysql, postgres, oracle, sql server) allow FOR UPDATE OF syntax.

long2ice: By researching, mysql, postgres, oracle support, and oracle is little different, sql server and sqlite is not support, so should I move that to dialects?

x8lucas8x: Can you please share your references (e.g. documentation links) for each one of them? Also summarising what are the main differences between the databases can help us to understand how to best implement FOR UPDATE OF support.

long2ice: - MySQL5.7, https://dev.mysql.com/doc/refman/5.7/en/select.html, don't support `of` and `nowait` and `skip locked`
- MySQL8, https://dev.mysql.com/doc/refman/8.0/en/select.html
- Postgres, https://www.postgresql.org/docs/current/sql-select.html
- SQLite, http://www.sqlite.org/draft/lang_select.html, don't support
- SQLServer, https://docs.microsoft.com/en-us/sql/t-sql/queries/select-transact-sql?view=sql-server-ver15, don't support
- Oracle, https://docs.oracle.com/cd/B19306_01/server.102/b14200/statements_10002.htm, support but has difference with others, not sure

x8lucas8x: @long2ice thanks for compiling those references. We actually don't care about some older version of a database, like in the mysql 5.7 case, not supporting something. So ignoring it, it becomes clear that this functionality isn't standardised across databases. Oracle case, for instance, diverges only in FOR UPDATE OF, given it accepts stuff other than tables. Therefore differing from mysql and postgres, which are pretty much the same in this regard.

As a result of those discrepancies, I believe FOR UPDATE and extras should be implemented as part of the respective database dialects. The main difference is that your logic should be moved away from QueryBuilder, which is meant to only standard SQL, to one of the specific QueryBuilder subclasses on https://github.com/kayak/pypika/blob/master/pypika/dialects.py. Could you do that for both MySQLQueryBuilder (MySQLQueryBuilder) and PostgreSQLQueryBuilder (https://github.com/kayak/pypika/blob/master/pypika/dialects.py#L343)?

long2ice: OK, I will do that later.

gl3nn:
@long2ice Could you reply to the above comment? Would like to close this PR; ideally by merging.

x8lucas8x:
This was merged as part of a different PR (https://github.com/kayak/pypika/pull/618).

PR #566, opened 2021-03-07
Render subquery function arguments with parentheses

uhrm:
When using correlated subqueries as function arguments, they should be rendered with parentheses.

An example where this is required with JSON functions:
```sql
SELECT json_object(
    'id', T.id,
    'child', (
        SELECT json_object(
            'child_id', C.id,
            'child_attr', C.attr
        )
        FROM child C
        WHERE C.parent_id = T.id
    )
)
FROM table T
WHERE T.id = ?
```

### DIFF
--- a/pypika/terms.py
+++ b/pypika/terms.py
@@ -1246,7 +1246,7 @@ class Function(Criterion):
 
     @staticmethod
     def get_arg_sql(arg, **kwargs):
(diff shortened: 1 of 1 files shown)


x8lucas8x (changes requested):
Can we please add tests?

uhrm:
I noticed this is a duplicate of #549 which already provides the same bugfix.

Feel free to merge that request and disregard this one.

x8lucas8x:
I ended up merging the other PR.

PR #581, opened 2021-04-06
Feat/add regexp sqlite

### LINKED ISSUE #564: Snowflake REGEXP function supported?
Hi,

Is the [Snowflake REGEXP function](https://docs.snowflake.com/en/sql-reference/functions/regexp.html) supported?  I see a `regex` function in `pypika`:

~~~python
customers = Tables('customers')
q = Query.from_(customers).select(
    customers.id,
    customers.fname,
    customers.lname,
).where(
    customers.lname.regex(r'^[abc][a-zA-Z]+&')
)
~~~

Which creates:

~~~sql
SELECT id,fname,lname FROM customers WHERE lname REGEX '^[abc][a-zA-Z]+&';
~~~

Is there a way with a `dialect` to change REGEX to REGEXP?

thx

lyz-code:
Closes #564

The changes in the order of the imports were done by `isort`, if you want I can restore the order.

Also changed the version of black from `stable` to `master` to fix the CI

### DIFF
--- a/.github/workflows/black.yml
+++ b/.github/workflows/black.yml
@@ -1,3 +1,4 @@
+---
 name: Lint
 
 on: [push, pull_request]
@@ -8,4 +9,4 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
-      - uses: psf/black@stable
+      - uses: psf/black@master
--- a/pypika/enums.py
+++ b/pypika/enums.py
@@ -32,6 +32,7 @@ class Matching(Comparator):
     ilike = " ILIKE "
     rlike = " RLIKE "
     regex = " REGEX "
+    regexp = " REGEXP "
     bin_regex = " REGEX BINARY "
     as_of = " AS OF "
     glob = " GLOB "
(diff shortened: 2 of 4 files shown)


.github/workflows/black.yml
```diff
@@ -8,4 +9,4 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
-      - uses: psf/black@stable
+      - uses: psf/black@master
```
gl3nn, line 12: Why is this changed?

gl3nn: Oh because of "Error: Unable to resolve action `psf/black@stable`, unable to find version `stable`". I think it would be better to pin a version instead of master? Not sure.

lyz-code: Because black changed their main branch and broke all the CIs of the world.

Check the output of the job of commit a5931d7  -> https://github.com/kayak/pypika/runs/2277687520

gl3nn: Yeah I noticed. If using master is the recommended approach then I'm fine with it.

gl3nn (approved):

PR #583, opened 2021-04-08
Added LIMIT BY statement

nibrag:


x8lucas8x:
Is this ANSI SQL? Can you please share references of major DBS (e.g. mysql, postgres, sqllite, oracle) supporting this? If that's not standard, it should be defined as part of dialects.

PR #586, opened 2021-04-14
Added NOT GLOB operator

kapily:
According to:
https://sqlite.org/lang_expr.html

> Both GLOB and LIKE may be preceded by the NOT keyword to invert the sense of the test

In a previous PR, I had added `GLOB`. In this one, I'm adding `NOT GLOB`

### DIFF
--- a/pypika/enums.py
+++ b/pypika/enums.py
@@ -34,6 +34,7 @@ class Matching(Comparator):
     bin_regex = " REGEX BINARY "
     as_of = " AS OF "
     glob = " GLOB "
+    not_glob = " NOT GLOB "
 
 
 class Boolean(Comparator):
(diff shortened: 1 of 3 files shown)


elod-jpg:
2880

x8lucas8x:
Is this ANSI SQL? I couldn't find this being supported in Mysql (https://dev.mysql.com/doc/refman/8.0/en/string-comparison-functions.html#operator_like). If that's sqllite specific it and its counterpart should be part of the sqllite dialect. Not in terms.py.

x8lucas8x (changes requested):
Move non ANSI to the proper dialect.

x8lucas8x:
No, we can't leave it here.

kapily:
Can you give a pointer to the file where it should be moved to?

x8lucas8x:
@kapily definitely. It should be on https://github.com/kayak/pypika/blob/master/pypika/dialects.py#L724. Please move the original GLOB operator too, given it is also sql lite specific. The idea with dialects is that you are meant to use the dialect directly, if in need of a specific db functionality. Tests for both GLOB/ NOT GLOB should be moved to https://github.com/kayak/pypika/blob/master/pypika/tests/dialects/test_sqllite.py.

PR #589, opened 2021-04-23
Test tabled fields dict key collision

noambloom:
`pypika.Field` inherits its `__hash__` from `Term` defined as 
```python
hash(self.get_sql(with_alias=True))
```
However for tabled `Fields` e.g.
```python
fld1 = Field("name", table=Table("tname1"))
fld2 = Field("name", table=Table("tname2"))
fld1.get_sql(with_alias=True) == fld2.get_sql(with_alias=True) == "name"
```
And thus the `__hash__` function would return the same result making Field not suited for being a key in a dictionary.

### DIFF
--- a/pypika/tests/test_terms.py
+++ b/pypika/tests/test_terms.py
@@ -16,15 +16,27 @@ class FieldAliasTests(TestCase):
 
 
 class FieldHashingTests(TestCase):
-    def test_tabled_fields_dict_has_no_collisions(self):
+    def test_tabled_eq_fields_equally_hashed(self):
+        client_name1 = Field(name="name", table=Table("clients"))
+        client_name2 = Field(name="name", table=Table("clients"))
+        self.assertTrue(hash(client_name1) == hash(client_name2))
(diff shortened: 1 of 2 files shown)


pypika/tests/test_terms.py
```diff
@@ -15,6 +15,18 @@ def test_when_alias_specified(self):
         self.assertEqual('bar', str(c1.alias))
 
 
+class FieldHashingTests(TestCase):
```
x8lucas8x, line 18: Can you please add tests for when namespace isn't set also? Besides, can you please change your test to check whether the hash of field A is different from the hash of field B? Currently you are doing that in a very indirect way by checking the length of a dict, which have keys as fields. That definitely an over-engineered test, given you could just check whether the hash of them is equal/different.

x8lucas8x (changes requested):

gl3nn (approved):

x8lucas8x (approved):

PR #601, opened 2021-05-22
Improvement on TOP statement support for `mssql`.

jelther:
Added support for WITH TIES and PERCENT statements when using MSSQL dialect.


gl3nn (commented):
Please fix the linter. Also, would it make sense to have this TOP for other database engines as well?

jelther:
> Please fix the linter. Also, would it make sense to have this TOP for other database engines as well?

done. applied the black formatting (guess you meant that instead of pylinting the project).
The TOP clause is used exclusively on mssql, never heard of this structure on other dialects (https://www.w3schools.com/sql/sql_top.asp).

PR #604, opened 2021-06-12
Drop unnecessary .format()

shadchin:


gl3nn (commented):
Please fix the linter issues.

gl3nn (commented):
The black formatter still fails on this.

PR #631, opened 2021-08-29
fix: add _use/_force_indexes into __copy__ with shallow copy

elisong:
By #334, #470 , pypika got `force_index`、`use_index` feature, but forgot to add "_force_indexes", "_use_indexes" into `QueryBuilder.__copy__()` with shallow copy, thus they both become mutable. For examples

```python
from pypika import Query

query = Query.from_('customers').select('id', 'fname')
q1 = query.use_index('idx1')
q2 = query
q3 = query.use_index('idx2')

if __name__ == "__main__":
    print(q1.get_sql())
    # SELECT "id","fname" FROM "customers" USE INDEX ("idx1","idx2")
    print(q2.get_sql())
    # SELECT "id","fname" FROM "customers" USE INDEX ("idx1","idx2")
    print(q3.get_sql())
    # SELECT "id","fname" FROM "customers" USE INDEX ("idx1","idx2")
```

### DIFF
--- a/pypika/queries.py
+++ b/pypika/queries.py
@@ -752,6 +752,8 @@ class QueryBuilder(Selectable, Term):
         newone._joins = copy(self._joins)
         newone._unions = copy(self._unions)
         newone._updates = copy(self._updates)
+        newone._force_indexes = copy(self._force_indexes)
+        newone._use_indexes = copy(self._use_indexes)
         return newone
 
     @builder
--- a/pypika/tests/test_immutability.py
+++ b/pypika/tests/test_immutability.py
@@ -1,17 +1,84 @@
 import unittest
 
 from pypika import (
+    AliasedQuery,
     Query,
     Tables,
 )
 
 
 class ImmutabilityTests(unittest.TestCase):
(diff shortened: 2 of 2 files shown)


x8lucas8x (changes requested):
@elisong that looks good, but can you please add tests to ImmutabilityTests, so that this can be tested. So far we only have join tests there (i.e. test_queries_after_join). If you could add test cases for other things that are being copied as well (see __copy__), such as selects, groupbys, and so on, that would be great (e.g. test_queries_after_groupby, test_queries_after_select).

x8lucas8x:
@elisong that looks good, but can you please add tests to ImmutabilityTests, so that this can be tested. So far we only have join tests there (i.e. test_queries_after_join). If you could add test cases for other things that are being copied as well (see __copy__), such as selects, groupbys, and so on, that would be great (e.g. test_queries_after_groupby, test_queries_after_select).

elisong:
@x8lucas8x Fine, I added some tests.

x8lucas8x:
@elisong thank you. It looks great now.

PR #643, opened 2021-10-08
add getitem notation

tobymao:
presto, hive, spark all use bracket notation for map access

field['foo'] etc


AzisK:
Thank you for your work. Could you please update from the main branch?

tobymao:
@AzisK no wonders, I get it.

Issue #579, opened 2021-03-23
Postgres UPDATE statements should convert lists to ARRAY


vfonte91:
Hello!

When using the Postgres dialect, a `list` isn't being converted to an `ARRAY` for `UPDATE` statements. Lists are handled correctly for `INSERT` statements and there's test coverage.

To replicate:

```python
from unittest.case import TestCase

from pypika import Table, Query, PostgreSQLQuery
from pypika.enums import Dialects

class UpdateTest(TestCase):

    def test_dialect(self):
        
        q = Query.update(Table("abc"), dialect=Dialects.POSTGRESQL).set(1, [1, "a", True])

        self.assertEqual("UPDATE \"abc\" SET \"1\" = ARRAY[1,'a',true]", str(q))

    def test_postgresquery(self):

        q = PostgreSQLQuery.update(Table("abc")).set(1, [1, "a", True])

        self.assertEqual("UPDATE \"abc\" SET \"1\" = ARRAY[1,'a',true]", str(q))

```
The resulting query is `UPDATE "abc" SET "1"=[1, 'a', True]` which is incorrect syntax.

Thanks!

foxx:
Just ran into this issue too, bumping.

PR #644, opened 2021-10-17
Fix Postgres array syntax for UPDATE

raxod502:
This PR should address https://github.com/kayak/pypika/issues/579 by formatting `UPDATE` queries using the same rules as `INSERT` queries. Regression tests for the new case have been added, courtesy of @vfonte91.

A side effect of this fix is that I discovered PyPika has inconsistent handling of `null` versus `NULL`. In these queries, `null` is used:

```
SELECT * FROM "abc" WHERE "foo"=null
UPDATE "abc" SET "foo"=null
```

But in these queries, `NULL` is used:

```
INSERT INTO "abc" VALUES (NULL)
INSERT INTO "abc" VALUES (1) RETURNING NULL
SELECT NULL "abcdef"
```

This is because there are two different code paths for formatting null values:

https://github.com/kayak/pypika/blob/a7b01dac790f4c6e93a961b0be26a9645fe7cf5c/pypika/terms.py#L385
https://github.com/kayak/pypika/blob/a7b01dac790f4c6e93a961b0be26a9645fe7cf5c/pypika/terms.py#L479

Both of these code paths used to return lowercase `null` circa 2018; this was changed in a seemingly unrelated commit https://github.com/kayak/pypika/commit/3bfee98bb226d4736b03473137e45e4d3e78a127.

Some other keywords (`true`, `false`) seem to be lowercase by convention in PyPika, whereas others (`SYSTEM_TIME`) seem to be uppercase by convention, so it's not entirely clear to me what the expected standardization is. On the plus side, since SQL is fully case-insensitive, it doesn't really matter too much.

In this PR, I've updated all code paths and tests to use lowercase `null`. Let me know if that's wrong.

### DIFF
--- a/pypika/queries.py
+++ b/pypika/queries.py
@@ -1070,7 +1070,9 @@ class QueryBuilder(Selectable, Term):
     @builder
     def set(self, field: Union[Field, str], value: Any) -> "QueryBuilder":
         field = Field(field) if not isinstance(field, Field) else field
-        self._updates.append((field, self._wrapper_cls(value)))
+        if not isinstance(value, Term):
+            value = self.wrap_constant(value, wrapper_cls=self._wrapper_cls)
+        self._updates.append((field, value))
 
     def __add__(self, other: "QueryBuilder") -> _SetOperation:
         return self.union(other)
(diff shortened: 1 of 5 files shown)


levintrix:
would love this to be merged :(

foxx:
@ Maintainers - any chance of getting this merged?

raxod502:
@AzisK How's this look?

PR #662, opened 2022-01-03
Add logarithm functions

cpinamtz:


AzisK (approved):
It looks good. Thanks for working on this. Could update the code with the newest code from master branch so that tests would run?

PR #688, opened 2022-07-01
feat:add-functions

YX-XiaoBai:
now is not support functions, support sql like
```python
...
from pypika import Count
Query.from_(table).select(Count(*))
...
```


AzisK:
Does it make sense to have a test for this?

Issue #366, opened 2020-01-08
Strings with trailing backslashes generate invalid SQL


stickystyle:
(Might be related to #238)

If you have a string value that ends in an escaped backslash, the query generated will be invalid because it will escape the quoted string in the query.

```python
>>> import pypika
>>> 
>>> table = pypika.Table("a_table")
>>> query = pypika.MySQLQuery.update(table).set(table.column_1, "value_1\\")
>>> print(str(query))
UPDATE `a_table` SET `column_1`='value_1\'
```

Cruuncher:
This one is a real gotcha that wouldn't pop up in testing either, and bite you in production

p-baleine:
I think that Parametrized Queries are utilizable for in such cases.

https://pypika.readthedocs.io/en/latest/2_tutorial.html#parametrized-queries

Issue #238, opened 2019-03-28
Enhance escape string


StevenKangWei:
Currrent escape have some issue for insert statement, some improve like below.

```python
class Value(Term):

    def __init__(self, value=None):
        self.value = value

    def get_sql(self, quote_char=None, **kwargs):
        import json
        import pymysql
        from datetime import date

        if isinstance(self.value, dict):
            return "'{}'".format(pymysql.escape_string(json.dumps(self.value)))
        if isinstance(self.value, date):
            return "'{}'".format(self.value.isoformat())
        if isinstance(self.value, str):
            return "'{}'".format(pymysql.escape_string(self.value))
        if isinstance(self.value, bool):
            return str.lower(str(self.value))
        if self.value is None:
            return 'null'
        return str(self.value)
```

twheys:
Not sure what issues you're talking about. Would you maybe raise a PR with some test cases?

StevenKangWei:
If some string include '\\' or another special case in insert or update statement that SQL will be unexpected. there is a potential SQL injection.

pierresouchay:
Hello,

I have really weird stuff on my side, with errors on SQL when values such as `"{"hello": "it's time"}` are sent (with a database using "'" as separator).

Okay, the code is properly not perfect, but I am very puzzled by https://github.com/kayak/pypika/blob/master/pypika/terms.py#L370 => "# FIXME escape values"

=> On my side, I see some of our developers using pypika and creating SQL Injections... Maybe we are just dumb, but it looks pypika help people creating SQL injections.

In this case, It was some kind of JSON value, but the way it is handled in the code by defaulting to str(value) looks very scary to me... I would expect something more exhaustive to avoid creating SQL injections.

Not to mention https://github.com/kayak/pypika/issues/366 which is very scary as well

PR #689, opened 2022-07-05
fix for better escaping to avoid SQL injections

pierresouchay:
Fix https://github.com/kayak/pypika/issues/238

Fix https://github.com/kayak/pypika/issues/366

### DIFF
--- a/pypika/terms.py
+++ b/pypika/terms.py
@@ -353,6 +353,17 @@ class Negative(Term):
         return "-{term}".format(term=self.term.get_sql(**kwargs))
 
 
+def escaped_param_str(value: str, quote_char: str) -> str:
+    # Did not find how to do that simply with pypika :(
+    return (
+        value.replace("\\", r"\\")
+        .replace(quote_char, quote_char * 2)
+        .replace("\r", r"\r")
+        .replace("\n", r"\n")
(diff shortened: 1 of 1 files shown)


AzisK:
Could you write tests for this?

pypika/terms.py
```diff
@@ -353,6 +353,17 @@ def get_sql(self, **kwargs: Any) -> str:
         return "-{term}".format(term=self.term.get_sql(**kwargs))
 
 
+def escaped_param_str(value: str, quote_char: str) -> str:
+    # Did not find how to do that simply with pypika :(
+    return (
+        value.replace("\\", r"\\")
+        .replace(quote_char, quote_char * 2)
```
gavindsouza, line 360: To add something like this now, it should be configurable behind a feature flag. This gives off big breaking energy 🥲

AzisK:
Please also pull the newest code from master for the CI tests to pitch in

pypika/terms.py
pierresouchay: OK, why not, how do you see it. a default option set to escape function by default on which you can override and put a no-op one? 

I am here because I saw some SQL injections on our platform, IMHO, it should be applied by default

PR #691, opened 2022-07-12
Added quote chars to 'with' aliases

mcsj120:


mcsj120:
@AzisK Only took a year and a half, but rebased :)

PR #697, opened 2022-08-06
Fix Type Hint

thislight:
This pull request is intended to revised existing type hint in this project and fix all problems about typing reported by [mypy](https://mypy.readthedocs.io).

Including fixes:
- `__eq__()` and `__ne__` accept any object as the argument.
- correct type hint for decorator `@builder`

### DIFF
--- a/pypika/dialects.py
+++ b/pypika/dialects.py
@@ -105,14 +105,14 @@ class MySQLQueryBuilder(QueryBuilder):
     @builder
     def for_update(
         self, nowait: bool = False, skip_locked: bool = False, of: TypedTuple[str, ...] = ()
-    ) -> "QueryBuilder":
+    ):
         self._for_update = True
         self._for_update_skip_locked = skip_locked
         self._for_update_nowait = nowait
         self._for_update_of = set(of)
 
     @builder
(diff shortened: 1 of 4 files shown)


thislight:
I discover a failed test for terms.py:
````
ERROR: test__criterion_replace_table_with_value (pypika.tests.test_internals.TablesTests)
----------------------------------------------------------------------
  File "~/Project/pypika/pypika/tests/test_internals.py", line 16, in test__criterion_replace_table_with_value
    self.assertEqual(c.left, table)
  File "/usr/lib64/pypy3.9/unittest/case.py", line 837, in assertEqual
    assertion_func(first, second, msg=msg)
  File "/usr/lib64/pypy3.9/unittest/case.py", line 827, in _baseAssertEqual
    if not first == second:
  File "~/Project/pypika/pypika/terms.py", line 265, in __eq__
    return BasicCriterion(Equality.eq, self, Term._assert_guard(self.wrap_constant(other)))
  File "~/Project/pypika/pypika/terms.py", line 301, in _assert_guard
    raise TypeError("expect Term object, got {}".format(type(v).__name__))
TypeError: expect Term object, got Table
````
I wonder if the original behaviour in the test is intended.

Issue #533, opened 2020-12-20
Insert dictionary


jly36963:
I wish their was a built-in way to insert dictionaries.

This is what I wish existed

```py
from pypika import Query, Table

users = Table('users')

Query.into(users).insert_dict({
    'first_name': 'John',
    'last_name': 'Doe',
})
```

This is what I'm currently doing:

```py
@classmethod
def insert_dict(self, table, d): return (
    self
    .into(table)
    .columns(*d.keys())
    .insert(*d.values())
)

Query.insert_dict = insert_dict

Query.insert_dict(users, {
    'first_name': 'John',
    'last_name': 'Doe',
})

PR #703, opened 2022-09-14
Easier inserts

charlie572:
I've added a feature that allows you to insert values from a dictionary, using the keys as the column names. This addresses #533. I've also added a feature that allows you to update all the given values when there is a duplicate key.


charlie572:
I noticed that bytes values don't get formatted correctly. I've fixed this and added it to this pull request.

charlie572:
I probably should have put these in separate pull requests. Sorry.

PR #707, opened 2022-10-12
Add support of SAMPLE ClickHouse clause

YiuRULE:


AzisK:
Could you pull from the master branch so that the newest CI test kick in?

YiuRULE:
done, it has been rebase to master

pypika/dialects.py
```diff
@@ -781,6 +781,16 @@ def drop_view(self, view: str) -> "ClickHouseDropQueryBuilder":
 class ClickHouseQueryBuilder(QueryBuilder):
     QUERY_CLS = ClickHouseQuery
 
+    def __init__(self, **kwargs) -> None:
+        super().__init__(self, as_keyword=True)
```
AzisK, line 785: is there no need to send other `kwargs` to the `super().__init__`?

YiuRULE: I think my previous code was indeed wrong, it has been corrected on the latest revision, thanks for bringing up the issue

AzisK (approved):
I have not had a chance to use Clickhouse in my life but the code looks good. Thanks for your work

PR #708, opened 2022-10-15
Adding alias to PseudoColumn

spongini:


Ga68:
you probably want an extra if-statement in there, like in here

https://github.com/kayak/pypika/pull/735

to handle when the pseudo column is used outside of the select statement (ex. in order-by or group-by)

PR #725, opened 2023-04-18
simple insert_dict() builder

edvardm:
## Additions

- add helper `insert_dict()` for QueryBuilder instances
- guard to prevent use of `columns()` while using `insert_dict()`
- basic tests

Idea pretty much the same as in issue #533 :

```python
Query().into(self.table_abc).insert_dict({"num": 1, "timestamp": datetime(2023, 4, 18)})

# INSERT INTO "abc" ("num","timestamp") VALUES (1,'2023-04-18T00:00:00')
```

Given how it's implemented, maybe `from_dict()` would be more suitable, as `into()` is already part of context? Multi-value inserts are trickier though, as they seem to vary between engines

----

Edit:

- Simple example in README
- Just use class method `Query.into` in tests

### DIFF
--- a/pypika/tests/test_inserts.py
+++ b/pypika/tests/test_inserts.py
@@ -169,6 +169,7 @@ class InsertIntoTests(unittest.TestCase):
             'WITH sub_qs AS (SELECT "id" FROM "abc") INSERT INTO "abc" SELECT "sub_qs"."id" FROM sub_qs', str(q)
         )
 
+
 class InsertIntoWithDict(unittest.TestCase):
     table_abc = Table("abc")
 
@@ -183,7 +184,10 @@ class InsertIntoWithDict(unittest.TestCase):
     def test_inserting_dictionary_produces_builder(self):
         q = Query.into(self.table_abc).insert_dict({"num": 1, "timestamp": datetime(2023, 4, 18)})
         q = q.insert(2, datetime(2023, 4, 19))
-        self.assertEqual("INSERT INTO \"abc\" (\"num\",\"timestamp\") VALUES (1,'2023-04-18T00:00:00'),(2,'2023-04-19T00:00:00')", str(q))
+        self.assertEqual(
+            "INSERT INTO \"abc\" (\"num\",\"timestamp\") VALUES (1,'2023-04-18T00:00:00'),(2,'2023-04-19T00:00:00')",
+            str(q),
+        )
 
     def test_columns_is_not_allowed_with_insert_dict_even_with_matching_columns(self):
         with self.assertRaisesRegex(QueryException, "Cannot mix use of columns.*and insert_dict"):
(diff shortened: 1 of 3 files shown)


pypika/tests/test_inserts.py
```diff
@@ -168,6 +169,28 @@ def test_insert_with_statement(self):
             'WITH sub_qs AS (SELECT "id" FROM "abc") INSERT INTO "abc" SELECT "sub_qs"."id" FROM sub_qs', str(q)
         )
 
+class InsertIntoWithDict(unittest.TestCase):
+    table_abc = Table("abc")
+
+    def test_inserting_simple_dictionary(self):
+        q = Query().into(self.table_abc).insert_dict({"c1": "value", "c2": 1})
```
wd60622, line 176: Any reason for not using `Query.into` since `into` is a classmethod?

pypika/tests/test_inserts.py
```diff
@@ -168,6 +169,28 @@ def test_insert_with_statement(self):
...
+        self.assertEqual("INSERT INTO \"abc\" (\"num\",\"timestamp\") VALUES (1,'2023-04-18T00:00:00')", str(q))
+
+    def test_inserting_dictionary_produces_builder(self):
+        q = Query().into(self.table_abc).insert_dict({"num": 1, "timestamp": datetime(2023, 4, 18)})
+        q = q.insert(2, datetime(2023, 4, 19))
+        self.assertEqual("INSERT INTO \"abc\" (\"num\",\"timestamp\") VALUES (1,'2023-04-18T00:00:00'),(2,'2023-04-19T00:00:00')", str(q))
+
+    def test_columns_is_not_allowed_with_insert_dict(self):
+        with self.assertRaises(QueryException):
+            Query().into(self.table_abc).columns("a", "b").insert_dict({"num": 1})
```
wd60622, line 190: Any more to be checked about the error? Maybe regex match on the message?

i.e https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaisesRegex

williambdean (commented):
Able to add this into the README in order to document?

pypika/tests/test_inserts.py
edvardm: These are all good and valid points. Not working that much Python nowadays, but should be simple address. Thanks!

edvardm:
Sorry for delay, some quite tragic happened at work and forgot about this for some time. Hopefully resolved noted issues

AzisK:
Hey @edvardm. I am sorry I am looking at this only now. Is it still relevant?

edvardm:
NP, it was  just what I had found useful at the time in a small internal tool -> closing

AzisK:
It seems like a neat addition but up to you @edvardm

PR #733, opened 2023-07-07
Vertica alias fix

Ga68:
Vertica requires **no** aliases in the `GROUP BY` and `ORDER BY` clauses. The current implementation does not do that.


Ga68:
incorrectly branched, closing request

PR #734, opened 2023-07-07
Vertica alias use fix

Ga68:
Vertica requires no aliases in the GROUP BY and ORDER BY clauses. The current implementation does not do that.

### DIFF
--- a/pypika/dialects.py
+++ b/pypika/dialects.py
@@ -258,6 +258,8 @@ class VerticaQueryBuilder(QueryBuilder):
         self._hint = label
 
     def get_sql(self, *args: Any, **kwargs: Any) -> str:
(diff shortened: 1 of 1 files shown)


AzisK:
Could you point me to the documentation and also add a test for this feature? Thank you for your work

Ga68:
I'm glad you asked about the documentation. Turns out I was wrong about this one. I can't remember at this point what made me think it was needed, but checking here

https://docs.vertica.com/12.0.x/en/sql-reference/statements/select/#parameters

it says

> You can optionally assign a temporary alias to each column expression and reference that alias elsewhere in the SELECT statement—for example, in the query predicate or ORDER BY clause

Going to close this one as it seems un-needed.

PR #735, opened 2023-07-07
with_alias handling missing in various criterion

Ga68:
in many of the Criterion classes, the `with_alias` handling was missing or inconsistent


AzisK:
Thank you for the improvements. Could you please add tests for this functionality so that we are always certain in the future that specifying and not specifying the alias works correct?

PR #740, opened 2023-08-10
fix: Typing hint for builder decorator

gavindsouza:
Maintain that the Callable type received is the same as the Callable returned.

**Additional Context for the screenshots:** tbl is a `Table` object, `frappe.qb` is a namespace that has type `MySQLQuery | PostgreSQLQuery`

### Before

before-typing-change

### After

after-typing-change

### DIFF
--- a/pypika/utils.py
+++ b/pypika/utils.py
@@ -1,4 +1,4 @@
-from typing import Any, Callable, List, Optional, Type
+from typing import Any, Callable, List, Optional, Type, TypeVar
 
 __author__ = "Timothy Heys"
 __email__ = "theys@kayak.com"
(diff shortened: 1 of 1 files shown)


AzisK:
Could you elaborate on this change a little bit more? I don't fully get it

AzisK:
Interesting. Okay, thanks for the explanation as well as for your work

AzisK (approved):
Looks good

AzisK:
We need to lint the code as well as fix the unit test runners

gavindsouza:
@AzisK We should add a pre-commit config that "black"en's the codebase automatically by the way.

--

As for the failing 3.6 build, it's high time we drop it given that it's unmaintained and there are 6 new major versions released meanwhile - https://github.com/kayak/pypika/issues/747

AzisK:
I agree about the pre-commit hook. I just created an issue https://github.com/kayak/pypika/issues/748

I believe we can drop support for Python3.6, I am just not sure what to do with the version of the library. Do we have to bump the major version in this case?

AzisK:
@gavindsouza could you update this PR with the newest CI code from master?

gavindsouza:
> @gavindsouza could you update this PR with the newest CI code from master?

Done 😄

AzisK:
Good job. I am merging this but new version and release will be coming later

PR #742, opened 2023-08-16
fix `Field.__init__`

WH-2099:
According to the signature of `Field.__init__`, the argument `table` can be of type `str`. Therefore, we should also check the type and ensure that `self.table` is assigned a `Table` object.


pypika/terms.py
```diff
@@ -539,6 +539,8 @@ def __init__(
     ) -> None:
         super().__init__(alias=alias)
         self.name = name
+        if isinstance(table, str):
```
AzisK, line 542: Shall we also type-hint `Table` for the `table` variable? I am not sure what is this `Selectable` data type

AzisK: If we do not want `str`, then it is better to also remove it from type-hinting

AzisK (changes requested):
Looks good but I left a few comments. Could you take a look?

pypika/terms.py
WH-2099: According to /pypika/queries.py#L116, `Table` is a subclass of `Selectable`.
```python
class Table(Selectable):
    ...
```
So I think type-hint `Selectable` is enough.

WH-2099: I think `str` is still useful and provides easy access in many cases.
On the other hand this would seem to be more consistent with methods such as `select`, as they also accept `str`.

AzisK: Thank you for elaborating, I agree

AzisK: I would like to ask one more thing. Could you write a test that accepts both data types and works as expected?

WH-2099: 当然可以，下周我会补充上这个测试。

WH-2099:
@AzisK Sorry for the delay. I've finished the test case.

WH-2099:
Can anyone else help review this PR? I'm still waiting.

AzisK (approved):
Looks good. Thanks for the work

Issue #537, opened 2021-01-06
SELECT statements are enclosed in a parenthesis when using UNION / INTERSECTION operators


mukund109:
Pypika encloses the `SELECT` statements in a parenthesis when using a UNION or INTERSECTION operator (haven't tried other set operations)

```python
>>> pypika.__version__
'0.40.0'
>>> q = Query.from_('t1').select('*') + Query.from_('t2').select('*')
>>> q.get_sql()
'(SELECT * FROM "t1") UNION (SELECT * FROM "t2")'
```
SQLite throws a syntax error for this. It doesn't allow the `SELECT` statements to be enclosed in parenthesis  - [see SQLite docs](https://sqlite.org/syntax/compound-select-stmt.html)

According to SQLite this is the correct way of doing it
```sql
SELECT * FROM "t1" UNION SELECT * FROM "t2"
```

jubele:
I am experiencing same issue while executing `SELECT` statements using `EXCEPT` operator.

```python3
>>>pypika.__version__
'0.48.0'
>>>from pypika import Query
>>>from pypika import Table
>>>t1 = Table('t1')
>>>q1 = Query.from_(t1).select('*').orderby(t1.r1)
>>>q2 = Query.from_(t1).select('*').where(t1.r1.lte(10))
>>>q_ex = q1.except_of(q2)
>>>q_ex.get_sql()
'(SELECT * FROM "t1" ORDER BY "r1") EXCEPT (SELECT * FROM "t1" WHERE "r1"<=10)'
```

TarantoolDB (SQL-101 syntax) throws also syntax error due to parenthesis enclosure of the statements `tarantool.error.DatabaseError: (184, "Syntax error at line 1 near '('")` 

Correct syntax in my case is same as @mukund109 proposed.
```sql
SELECT * FROM "t1" ORDER BY "r1" EXCEPT SELECT * FROM "t1" WHERE "r1"<=10
```

daidai21:
https://github.com/kayak/pypika/blob/7d0bc587014b0180dfa826f477a06592d8484d01/pypika/tests/test_joins.py#L708-L712

https://github.com/kayak/pypika/blob/7d0bc587014b0180dfa826f477a06592d8484d01/pypika/tests/test_joins.py#L1114-L1117

test shell:
```sh
docker pull nouchka/sqlite3
docker run -it --rm nouchka/sqlite3
```

test sql:
```sql

CREATE TABLE a ( id int );
[...]

PR #746, opened 2023-09-06
Fix: Don't wrap set operations in brackets in SQLite

nwrahnema:
From https://github.com/kayak/pypika/issues/537, SQLite doesn't support set operations with brackets. Luckily there already exists this flag to disable adding brackets to set operations - enable for `SQLLiteQueryBuilder`.


AzisK:
Could you also add a test for this feature?

PR #666, opened 2022-01-31
Fix: Add py.typed to root package

whardier:
I don't believe I need to add this to the pypika.clickhouse package.  Works well in vscode and along with mypy.

PR #757, opened 2023-10-02
Make pypika compatible with PEP 561

fedetibaldo:
My local `mypy` setup had a few complaints about this package. In particular, as there are no stubs for `pypika` available on `typeshed`, it lamented the missing typings. However, by looking at the source, it's clear that the type annotations are indeed available—they're just inlined.

To fix that, I added a `py.typed` file inside my `.venv/.../pypika` folder. It worked, so I'm replicating those changes upstream. My reference for the implementation was the [mypy documentation](https://mypy.readthedocs.io/en/stable/installed_packages.html#creating-pep-561-compatible-packages). I've never published a package, so this is a stab in the dark for me. Hopefully, I hit the target 😜

Cheers


AzisK:
Hi @fedetibaldo, this is already being solved in https://github.com/kayak/pypika/pull/666. I see that you added it in the `MANIFEST.in` file. I am not sure if we need it here, let's read and ask around

Issue #723, opened 2023-04-06
Add a `pipe` method


wd60622:
Think that it could be pretty convenient to add a `pipe` method. Any thoughts? I can create a PR if it makes sense

```python 
from pypika import Query
from pypika.queries import QueryBuilder


def filter_to_desired_rows(query: QueryBuilder, values_to_select) -> QueryBuilder:
    ...

def aggregated_statistics(query: QueryBuilder, groups) -> QueryBuilder: 
    ...
    

initial_query = Query.from_("my_table")
query = (
    initial_query
    .pipe(filter_to_desired_rows, values_to_select=...)
    .pipe(aggregated_statistics, groups=...)
)
```

PR #759, opened 2023-10-09
add pipe operator on QueryBuilder

wd60622:
pipe method like in pandas in order to do some chaining. Related to this issue [here](https://github.com/kayak/pypika/issues/723).


pypika/__init__.py
```diff
@@ -100,7 +100,7 @@
 
 __author__ = "Timothy Heys"
 __email__ = "theys@kayak.com"
-__version__ = "0.48.9"
+__version__ = "0.48.10"
```
wd60622, line 103: is this the expected version update?

AzisK: I suggest to bump the version separately and also start a `CHANGELOG.md` as well as describe all of the changes there

pypika/tests/test_query.py
```diff
@@ -204,3 +204,29 @@ def test_query_builders_have_reference_to_correct_query_class(self):
...
+            (count_group, ["test1", "test2"], {}),
+            (count_group, ["test1"], {}),
+            (where_clause, [1], {}),
+            (where_clause, [], {"num_days": 1}),
+        ]:
+            self.assertEqual(str(base_query.pipe(func, *args, **kwargs)), str(func(base_query, *args, **kwargs)))
+
+        self.assertEqual(
+            str(base_query.pipe(select).pipe(where_clause, num_days=1)),
+            str(select(where_clause(base_query, num_days=1))),
```
AzisK, line 231: I would lean for tests not to rely on any code of the library and simply have the expected string here. Having said that, this can stay but I would also check if it is equal to a desired string of a query. What do you think?

AzisK (commented):
It looks good but I left a comment about the library as well as versioning approach

AzisK:
Thanks for you work!

pypika/tests/test_query.py
wd60622: Makes sense I will check the expected string

pypika/__init__.py
wd60622: I reverted the version back to 0.48.9 So no version change. 

Have you guys considered using the automated release notes? https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes 
It'd be based on the PR labels. Might be easy way to automate 😄

AzisK:
Could you describe this possibility in `README.rst` as well?

wd60622:
There is a new "Chaining Functions" section in the Tutorial 
Screenshot 2023-10-16 at 20 05 32

wd60622:
Any other suggestions here?

AzisK:
Looks good to me. If no other suggestions emerge in the coming days, I will merge it

PR #765, opened 2023-10-18
support final keyword for ClickHouse dialect

YiuRULE:


AzisK:
Thank you for your work

PR #769, opened 2023-11-13
fix: remove 'ValueError' from return union for 'wrap_constant'

guacs:
This removes the `ValueError` from the union in the return type of `Term.wrap_constant`.


wd60622 (approved):
Seems good! Any thoughts @AzisK ?

AzisK:
Yes, you can make another one

PR #771, opened 2023-11-24
Escape quotes in identifiers

### LINKED ISSUE #768: Identifier quote escaping
SQL identifier names are not being escaped, thus enabling SQL injection attacks. 

Here is a minimal example to reproduce:
```python
import pypika

table = pypika.Table('my_table"--')
field = getattr(table, 'my_field"--')
builder = (
    pypika.Query.from_(table, dialect=pypika.Dialects.POSTGRESQL)
    .select(field)
    .where(table.name == "value'")
)

print(builder)
```

This code produces the following SQL, where single quotes are correctly being escaped but double quotes are not:
```sql
SELECT "my_field"--" FROM "my_table"--" WHERE "name"='value'''
```

alexandermalyga:
Fix #768


pypika/tests/test_terms.py
```diff
@@ -73,3 +73,25 @@ def test_passes_kwargs_to_field_get_sql(self):
             'FROM "customers" JOIN "accounts" ON "customers"."account_id"="accounts"."account_id"',
             query.get_sql(with_namespace=True),
         )
+
+
+class IdentifierEscapingTests(TestCase):
+    def test_escape_identifier_quotes(self):
+        customers = Table('customers"')
+        customer_id = getattr(customers, '"id')
+        email = getattr(customers, 'email"').as_('customer_email"')
```
wd60622, line 82: what happens with single quotes inside double quotes?

PR #779, opened 2023-12-10
Doc extend operators

nlassaux:
Many extensions define new operators, for example to compute various distances between geometries, vectors, etc. A useful example is `<->` being used by PostGIS.

Added the documentation explaining how to leverage `ArithmeticExpression` to extend pypika with it, right under the explanation about how to define a custom function. Having that doc section would have saved me quite some time actually.

Thanks @wd60622 for explaining the approach!

### DIFF
--- a/docs/4_extending.rst
+++ b/docs/4_extending.rst
@@ -86,20 +86,16 @@ For instance, consider a scenario where you want to use the ``<->`` operator in 
 
 .. code-block:: python
 
-    from pypika import CustomFunction, Field, Query
+    from pypika import Field, Query
     from pypika.terms import ArithmeticExpression
 
     # Wrapper class for the custom operator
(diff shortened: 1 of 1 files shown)


wd60622:
Looks awesome. Thanks for making the adjustment

Only thing I would suggest is adding a brief section on this in the README like with the Custom Functions section

AzisK (approved):
Looks good!

AzisK:
Thanks to both of you. I don't have any additional suggestions except writing a test to be sure that people actually can define their own custom operators

wd60622:
> Thanks to both of you. I don't have any additional suggestions except writing a test to be sure that people actually can define their own custom operators

Think that's fair. If someones going into the code base, I think ideally someone might be able to use a string instead of a small class.

Issue #773, opened 2023-11-28
MySQL queries should wrap set operations


pandaxtc:
The MySQLQuery dialect doesn't wrap set operations in parens, but UNION queries that have LIMITs and ORDERs in the subqueries fail without parens. [The documentation states](https://dev.mysql.com/doc/refman/8.0/en/parenthesized-query-expressions.html#:~:text=You%20cannot%20have%20a%20query%20block%20with%20a%20trailing%20ORDER%20BY%20or%20LIMIT%20without%20wrapping%20it%20in%20parentheses%20but%20parentheses%20may%20be%20used%20for%20enforcement%20in%20various%20ways%3A):
> You cannot have a query block with a trailing ORDER BY or LIMIT without wrapping it in parentheses...

For example, this fails:
```sql
SELECT 1 LIMIT 1 UNION SELECT 2 LIMIT 1;
```
but this does not:
```sql
(SELECT 1 LIMIT 1) UNION (SELECT 2 LIMIT 1);
```

pandaxtc:
The example you provided uses `MSSQLQuery` instead of `MySQLQuery`. Here's an example:
```py
from pypika.dialects import MySQLQuery

query = MySQLQuery.from_("table").select("id").union(MSSQLQuery.from_("another_table").select("id"))
query.get_sql() # or str(query)
```
results in
```
'SELECT `id` FROM `table` UNION SELECT `id` FROM `another_table`'
```

wd60622:
I misread that then. 
Here in the code, the `wrap_set_operation_queries` is set to False.
https://github.com/kayak/pypika/blob/8841520e906970d76c5ed81c7dd5d154f0d5259d/pypika/dialects.py#L90

Would you want to make a PR to change?

In the meantime, you might be able to set the `wrap_set_operation_queries` attribute of your query to True.

pandaxtc:
The temporary solution works, thanks. I opened a PR: https://github.com/kayak/pypika/pull/782

PR #782, opened 2023-12-21
Make MySQL queries wrap set operations with parens

pandaxtc:
Add parens around set operations for MySQL. Fixes https://github.com/kayak/pypika/issues/773. 

Note that I removed the tests for checking no parens for MySQL, since adding back the parens brings the behavior in line with the default. I assume the tests are for checking that the behavior is not default. Let me know if the tests should be added back.

### DIFF
--- a/pypika/dialects.py
+++ b/pypika/dialects.py
@@ -87,7 +87,7 @@ class MySQLQueryBuilder(QueryBuilder):
     QUERY_CLS = MySQLQuery
 
     def __init__(self, **kwargs: Any) -> None:
-        super().__init__(dialect=Dialects.MYSQL, wrap_set_operation_queries=False, **kwargs)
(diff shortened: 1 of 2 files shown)


wd60622:
I'm not fully versed on the test suite, but it might be best to test that the union now **has** parentheses instead of deleting the previous test. No harm in double checking

pandaxtc:
Gotcha, just added the tests back.

wd60622 (approved):
Awesome. With passing tests, looks good to me!

wd60622:
@AzisK Could you run the test suite?

danielenricocahall:
@AzisK I hope all is going well! Would we be able to get this in as well? Looks like a small, useful change to include in a release if one is anticipated soon. 😄

AzisK:
Looks good to me

Issue #628, opened 2021-08-26
SQLite Insert or Ignore


marcstreeter:
While looking into https://github.com/tortoise/tortoise-orm/issues/527 I came upon the question of where there is any way to do the SQLLite query

```SQL
INSERT OR IGNORE INTO "foo" VALUES (1, 'm')
```

it seems like the SQLLiteQueryBuilder [insert_or_replace](https://github.com/kayak/pypika/blob/920a4cdd99cbcd85d55af2d8d2c601d8bef8bfc2/pypika/dialects.py#L847-L850) is very similar to what is needed. In copying `insert_or_replace`  it depends on the QueryBuilder's [replace_sql](https://github.com/kayak/pypika/blob/920a4cdd99cbcd85d55af2d8d2c601d8bef8bfc2/pypika/queries.py#L1390-L1392
) method . Using this pattern I propose to add a `insert_or_ignore` method to the SQLLiteQueryBuilder like so

```python
class SQLLiteQueryBuilder(QueryBuilder):
    QUERY_CLS = SQLLiteQuery

    def __init__(self, **kwargs: Any) -> None:
        # ... all the same just concatenating the below line
        self._insert_or_ignore = False

    @builder
    def insert_or_ignore(self, *terms: Any) -> "SQLLiteQueryBuilder":
        self._apply_terms(*terms)
        self._ignore = True  # not sure if this is good to change, seems like it's used elsewhere so maybe not this line
        self._insert_or_ignore = True

    def _ignore_sql(self, **kwargs: Any) -> str:
        prefix = "INSERT OR " if self._insert_or_ignore else ""
        return prefix + super()._ignore_sql(**kwargs)

```

and then the additional changes to the base QueryBuilder like so

```python
  
class QueryBuilder(Selectable, Term):
   #  ...maybe somewhere close to the `def _replace_sql` I don't know
  def _ignore_sql(self, **kwargs: Any) -> str:
        return "IGNORE INTO {table}".format(
            table=self._insert_table.get_sql(**kwargs),
        )
```

this would then allow an interaction like so

```python
from pypika import SQLLiteQuery
SQLLiteQuery.into("foo").insert_or_ignore(1, 'm')
```
and render out to

```SQL
INSERT OR IGNORE INTO "foo" VALUES (1, 'm')
```
maybe this is already possible via other mechanisms, pypika is pretty sweet already so please correct me if I'm wrong.

### considerations

[...]

PR #788, opened 2024-01-24
Use correct "insert ignore" syntax for Sqlite.

mchaput:
This is a simple patch to fix issue #628.


wd60622:
Would you be able to test the change?

danielenricocahall (approved):

PR #792, opened 2024-02-21
:bug: Fixes #791

### LINKED ISSUE #791: Pass kwargs through to `get_function_sql`
kwargs are passed around everywhere except in `Function.get_sql`.

I've implemented a parameter wrapper to ensure I can build parametrized queries, but this doesn't work on functions because of this simple issue.

https://github.com/kayak/pypika/blob/master/pypika/terms.py#L1336

mvanderlee:


AzisK:
Looks good to me. @wd60622 could you take a look?

Could you also add tests?

AzisK:
What do you think @wd60622?

PR #794, opened 2024-02-22
:star: Improve parameterized query support - fixes #793

### LINKED ISSUE #793: Improved parameter support
We're using pypika to build queries dynamically based on API filter input. Since these can get very complex it's impossible to keep track of the parameter indices or names. 

It'd be great if PyPika could support this out of the box, perhaps based on my suggestion in 2021 https://github.com/kayak/pypika/pull/201#issuecomment-816029029

mvanderlee:
Fixes #793 
The implemented tests speak for themselves, but here is an extract:

```python
subquery = (
    Query.from_(self.table_efg)
    .select(self.table_efg.fiz, self.table_efg.buz)
    .where(self.table_efg.buz == 'buz')
)

q = (
    Query.from_(self.table_abc)
    .join(subquery)
    .on(self.table_abc.bar == subquery.buz)
    .select(self.table_abc.foo, subquery.fiz)
    .where(self.table_abc.bar == 'bar')
)

parameter = NamedParameter()
sql = q.get_sql(parameter=parameter)
self.assertEqual(
    'SELECT "abc"."foo","sq0"."fiz" FROM "abc" JOIN (SELECT "fiz","buz" FROM "efg" WHERE "buz"=:param1)'
    ' "sq0" ON "abc"."bar"="sq0"."buz" WHERE "abc"."bar"=:param2',
    sql,
)
self.assertEqual({'param1': 'buz', 'param2': 'bar'}, parameter.get_parameters())
```

This would allow the user to easily plug into SQLAlchemy for example:

```python
parameter = NamedParameter()
session.execute(text(q.get_sql(parameter=parameter)), **parameter.get_parameters())
```

### DIFF
--- a/pypika/terms.py
+++ b/pypika/terms.py
@@ -299,15 +299,12 @@ def named_placeholder_gen(idx: int) -> str:
 class Parameter(Term):
     is_aggregate = None
 
-    def __init__(self, placeholder: Union[str, int, Callable[[int], str]] = idx_placeholder_gen) -> None:
+    def __init__(self, placeholder: Union[str, int]) -> None:
         super().__init__()
         self._placeholder = placeholder
 
     @property
     def placeholder(self):
-        if callable(self._placeholder):
-            return self._placeholder(None)
-
         return self._placeholder
 
     def get_sql(self, **kwargs: Any) -> str:
@@ -322,8 +319,7 @@ class Parameter(Term):
 
 class ListParameter(Parameter):
     def __init__(self, placeholder: Union[str, int, Callable[[int], str]] = idx_placeholder_gen) -> None:
-        super().__init__()
-        self._placeholder = placeholder
+        super().__init__(placeholder=placeholder)
         self._parameters = list()
 
     @property
@@ -342,8 +338,7 @@ class ListParameter(Parameter):
 
 class DictParameter(Parameter):
     def __init__(self, placeholder: Union[str, int, Callable[[int], str]] = named_placeholder_gen) -> None:
-        super().__init__()
-        self._placeholder = placeholder
+        super().__init__(placeholder=placeholder)
         self._parameters = dict()
 
     @property
--- a/pypika/terms.py
+++ b/pypika/terms.py
@@ -3,7 +3,7 @@ import re
 import uuid
 from datetime import date
 from enum import Enum
-from typing import TYPE_CHECKING, Any, Iterable, Iterator, List, Optional, Sequence, Set, Type, TypeVar, Union
+from typing import TYPE_CHECKING, Any, Callable, Iterable, Iterator, List, Optional, Sequence, Set, Type, TypeVar, Union
 
 from pypika.enums import Arithmetic, Boolean, Comparator, Dialects, Equality, JSONOperators, Matching, Order
 from pypika.utils import (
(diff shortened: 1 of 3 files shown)


pypika/terms.py
```diff
@@ -288,57 +288,116 @@ def get_sql(self, **kwargs: Any) -> str:
...
-    def __init__(self, placeholder: Union[str, int]) -> None:
+    def __init__(self, placeholder: Union[str, int, Callable[[int], str]] = idx_placeholder_gen) -> None:
         super().__init__()
-        self.placeholder = placeholder
+        self._placeholder = placeholder
+
+    @property
+    def placeholder(self):
+        if callable(self._placeholder):
+            return self._placeholder(None)
```
wd60622, line 309: how does this work on None?

wd60622:
Thanks for the PR! Will take a deeper look later

pypika/terms.py
mvanderlee: callable Returns `False`
image

But I presume you mean how will the `self_placeholder(None)` work with None.
Good point, it wouldn't atm with the default generators.

I will fix it now.

pypika/terms.py
```diff
@@ -385,9 +439,55 @@ def get_formatted_value(cls, value: Any, **kwargs):
...
+            # Don't stringify numbers when using a parameter
+            if isinstance(self.value, (int, float)):
+                value_sql = self.value
+            else:
+                value_sql = self.get_value_sql(quote_char=quote_char, **kwargs)
+            param_sql = parameter.get_sql(**kwargs)
+            param_key = parameter.get_param_key(placeholder=param_sql)
+            parameter.update_parameters(param_key=param_key, value=value_sql, **kwargs)
+
+            return format_alias_sql(param_sql, self.alias, quote_char=quote_char, **kwargs)
```
wd60622, line 462: Nitpick: de-indent since the if clause returns

pypika/terms.py
```diff
@@ -385,9 +439,55 @@ def get_formatted_value(cls, value: Any, **kwargs):
...
+            # Don't stringify numbers when using a parameter
+            if isinstance(self.value, (int, float)):
+                value_sql = self.value
+            else:
+                value_sql = self.get_value_sql(quote_char=quote_char, **kwargs)
+            param_sql = self._parameter.get_sql(**kwargs)
+            param_key = self._parameter.get_param_key(placeholder=param_sql)
+            parameter.update_parameters(param_key=param_key, value=value_sql, **kwargs)
+
+            return format_alias_sql(param_sql, self.alias, quote_char=quote_char, **kwargs)
```
wd60622, line 490: This seems to be the same as the `get_sql` above. Could it be inherited somehow? Maybe wrapped up in a common function?

pypika/terms.py
```diff
@@ -288,57 +288,111 @@ def get_sql(self, **kwargs: Any) -> str:
...
+
+
+def named_placeholder_gen(idx: int) -> str:
+    return f'param{idx + 1}'
+
+
 class Parameter(Term):
     is_aggregate = None
 
     def __init__(self, placeholder: Union[str, int]) -> None:
```
wd60622, line 302: Type hint here doesn't include callable? Is that correct still?

pypika/terms.py
```diff
@@ -288,57 +288,111 @@ def get_sql(self, **kwargs: Any) -> str:
...
         super().__init__()
-        self.placeholder = placeholder
+        self._placeholder = placeholder
```
wd60622, line 304: Might be good to use a setter here. With that, implementing a check for the correct type can happen on initialization if put in the setter

wd60622 (commented):
Looks really good. 
Just some comments on type hints, raising errors, and combining logic

wd60622:
Do you have any thoughts as well? @AzisK

pypika/terms.py
mvanderlee: Correct, only ListParameter and DictParameter (and classes that inherit from them) support callable. 
The base parameter does not.

pypika/terms.py
mvanderlee: There is a slight difference, But I've broken it out into `_get_param_data`

pypika/terms.py
mvanderlee: I'm not sure I follow how to do this pythonically? Could you share some pseudo code to show what you mean?

mvanderlee:
@wd60622 @AzisK any update on this? 
It should go hand in hand with #792

pypika/terms.py
danielenricocahall: @mvanderlee I believe @wd60622 was referring to the following:
```python
@property
def placeholder(self):
      return self._placeholder

@placeholder.setter
def placeholder(self, placeholder):
     # can add checks here if needed
     self._placeholder = placeholder
```

mvanderlee: Right, which adds no value and isn't pythonic. This isn't Java/C#.

It would also break parameters. After a parameter has been created and used, any changes to the placeholder will break queries. Which is why I've made it _private_ with a getter only.

I haven't changed the type, so any type checks would have to be part of a separate PR.

danielenricocahall: I would disagree with you on it not being Pythonic, as it’s a native language feature. 

Also, I was trying to be helpful, but given your receptiveness to discussion and tone, I won’t proceed. 

Good luck.

mvanderlee: Sorry for my tone @danielenricocahall it'd been a rough week and I lashed out.

As for pythonic. This video does a good job at explaining language feature vs pythonic. 
Raymond Hettinger - Beyond PEP 8 -- Best practices for beautiful intelligible code - https://www.youtube.com/watch?v=wf-BqAjZb8M

AzisK:
Could we also have this functionality describe in pypika docs?

reasv:
I just wanted to say that I'm going to fork your branch because this feature is really useful, and I'm using it in my own project to help me build a custom high level query language: https://github.com/reasv/panoptikon/blob/master/src/panoptikon/db/pql/build_query.py

I'm resorting to making my own fork of PyPika and merging a bunch of these PRs because it doesn't seem PyPika is getting much merged lately and there are open bugs that otherwise block development: https://github.com/kayak/pypika/pull/746/commits/d7a40549707c23b6debff138034c14b0f5c4c590

larsschwegmann:
Any progress on getting this merged? :)

AzisK:
Hello, I asked @mvanderlee for a test for `PyformatParameter` and I believe it would be good to have it. Otherwise I could merge this PR

larsschwegmann:
I forked @mvanderlee 's fork and added a test for PyformatParameter: https://github.com/mvanderlee/pypika/compare/mvanderlee/parametized_query...larsschwegmann:pypika:mvanderlee/parametized_query

AzisK:
@larsschwegmann could you make a PR? Is it better for me to push this change to this commit? Probably

larsschwegmann:
If it's possible, I guess it would be best if you could just add the commit to this PR, just for the sake of documentation. Then you'll still have all the discussion in the PR that was merged 😄 
If not, I can create a PR. Let me know what works best for you.

mvanderlee:
Thank you @larsschwegmann I've added it to the branch!

mvanderlee:
The affected code wasn't in this fork yet, so I rebased in order to fix it.

PR #803, opened 2024-05-14
Add support of SETTINGS ClickHouse clause

martijnthe:
### Summary

Clickhouse SQL has an optional [`SETTINGS` clause][settings-clause] with `SELECT`s which can be used to configure [all kinds of options][settings].

This PR adds a `settings` method to `ClickHouseQueryBuilder` to add these options to a `SELECT` query.

All current (Clickhouse 24.4) settings keys are snake-case. Therefore I opted to use the settings keys directly as kwargs to the `settings` method, allowing passing multiple settings pairs in one call in a Pythonic/ideomatic way.

[settings-clause]: https://clickhouse.com/docs/en/sql-reference/statements/select#settings-in-select-query
[settings]: https://clickhouse.com/docs/en/operations/settings/settings

 ### Test Plan

Added a unit test to exercise the new function.

### DIFF
--- a/pypika/dialects.py
+++ b/pypika/dialects.py
@@ -798,11 +798,17 @@ class ClickHouseQueryBuilder(QueryBuilder):
         super().__init__(**kwargs)
         self._sample = None
         self._sample_offset = None
+        self._settings = None
 
     @builder
     def sample(self, sample: int, offset: Optional[int] = None) -> "ClickHouseQueryBuilder":
         self._sample = sample
         self._sample_offset = offset
+
+    @builder
+    def settings(self, **kwargs: Any) -> "ClickHouseQueryBuilder":
+        self._settings = self._settings.copy() if self._settings else {}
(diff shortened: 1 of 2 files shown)


martijnthe:
cc @AzisK -- tagging you because you reviewed a very similar PR (#707). Thanks for looking!

AzisK:
Hi @martijnthe. Did you close because you lost hope someone will review? I am sorry about it. I can find the time on the weekend to review

AzisK (commented):
It looks good to me

martijnthe:
I also closed it because there's a problem with the implementation:
- `SETTINGS` must come between `LIMIT` and `UNION`, not right after the `FROM` part of a `SELECT`.
- Other statement kinds (i.e. `INSERT`) can also take `SETTINGS`, but it goes in a different place.

I didn't see a clean way to implement it w/o touching the base `QueryBuilder` class too.

PR #814, opened 2024-08-24
fix: select arg type

trim21:
here `terms` is already `*terms` so there is no need to add `Sequence`. adding `Sequence` means `terms` is `Tuple[Sequence[...]]`。

And it will cause a type error when `query.select(Count("1"))`


AzisK:
Hey @trim21. I know that nobody responded to you and sorry about this but is this the reason why you closed this pull request or was there another reason?

trim21:
https://github.com/kayak/pypika/pull/815 include this

PR #815, opened 2024-08-24
fix some types when downstream enabled this package in mypy

trim21:
mypy raise some error when I add py.typed file manually (https://github.com/kayak/pypika/pull/666)

details:

add `pypika.__all__`

fix select command, `def select(self, *terms: Sequence[Union[int, float, str, bool, Term, Field]]) -> "QueryBuilder":` to `def select(self, *terms: Union[int, float, str, bool, Term, Field]) -> "QueryBuilder":`

add common types to subclass of `DistinctOptionFunction`

### DIFF
--- a/pypika/__init__.py
+++ b/pypika/__init__.py
@@ -112,9 +112,63 @@ from pypika.utils import (
     FunctionException,
 )
 
+
 __author__ = "Timothy Heys"
 __email__ = "theys@kayak.com"
 __version__ = "0.48.9"
 
 NULL = NullValue()
 SYSTEM_TIME = SystemTimeValue()
+
+__all__ = (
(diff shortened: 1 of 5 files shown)


AzisK:
Does it make sense to add tests about it?

trim21:
not really for this PR

AzisK:
It looks good to me. Thanks for the work. Merging

AzisK (approved):

AzisK:
Should I simply revert this or would you have some improvements regarding it?

I am not really a mypy user, therefore not my expertise

trim21:
Sorry, I don't use pypika anymore, you can consider revert it.

PR #817, opened 2024-09-12
Add DISTINCT ON and LIMIT BY support to ClickHouse dialect

bikeshedder:
This adds [`DISTINCT ON`](https://clickhouse.com/docs/en/sql-reference/statements/select/distinct) and [`LIMIT BY`](https://clickhouse.com/docs/en/sql-reference/statements/select/limit-by) support to the ClickHouse dialect.

I'm not too happy that I had to add the `LIMIT BY` to the `_apply_pagination` method and pass the `**kwargs` around but I don't see any less intrusive way of doing this at the moment.

### DIFF
--- a/pypika/dialects.py
+++ b/pypika/dialects.py
@@ -1,7 +1,7 @@
 import itertools
 import warnings
 from copy import copy
-from typing import Any, Optional, Union, Tuple as TypedTuple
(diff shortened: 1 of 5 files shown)


AzisK (approved):
Looks good to me

PR #823, opened 2024-12-23
Add a decorator for creating table via class-style.

YogiLiu:
This feature add a decorator to create table via class-style, it's friendly to type hint.

```python
from pypika import table_class, Table, Field, Query

@table_class('customers')
class Customer(Table):
    id = Field('id')
    first_name = Field('fname')
    last_name = Field('lname')
    phone = Field('phone')

q = Query.from_(Customer).select(Customer.id, Customer.first_name, Customer.last_name, Customer.phone)
```

The decorator's parameters are same as `Table.__init__`.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -107,7 +107,7 @@ The table also can create via class-style:
         last_name = Field('lname')
         phone = Field('phone')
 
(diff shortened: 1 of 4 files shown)


YogiLiu:
I think #824 is the better way to implement this feature.

PR #827, opened 2025-01-13
Add support of SETTINGS ClickHouse clause

martijnthe:
### Summary

Clickhouse SQL has an optional [`SETTINGS` clause](settings-clause) with
`SELECT`s which can be used to configure [all kinds of options](settings).

In #1 and #2, I took a stab at implementing this, but the implementation contained a bug:
string setting values need to be single-quoted. Also, some settings require a `Map` and
some require a `Bool`. Let's support these too.

[settings-clause]: https://clickhouse.com/docs/en/sql-reference/statements/select#settings-in-select-query
[settings]: https://clickhouse.com/docs/en/operations/settings/settings

 ### Test Plan

Updated the unit test to cover the different `SETTINGS` data types.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -3,7 +3,14 @@ PyPika - Python Query Builder
 
 .. _intro_start:
 
-|BuildStatus|  |CoverageStatus|  |Codacy|  |Docs|  |PyPi|  |License|
+|License|
+
+
+Fork
+----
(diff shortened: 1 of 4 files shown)


AzisK (commented):
It looks good

PR #828, opened 2025-02-06
Some fixes/improvements for sqlite

akhilnarang:


AzisK:
Thanks for your work. Could you please pull the code from the main branch as well as add tests for the logic you added?

AzisK:
Thank you for the work. Could you add tests for the cases you covered?

PR #864, opened 2026-01-30
typefix: update parameter types in aggregate and function classes to use Term

noambloom:
**What changed?**
Updated type annotations in `pypika/functions.py` for 28 function classes, changing parameter types from `str | Field` to `Term` (or `str | Term` for Count). Added `Term` import from `pypika.terms`.

Affected functions:
- Aggregate functions: Sum, Avg, Min, Max, Std, StdDev, Abs, First, Last
- Math functions: Sqrt, Floor, Signed, Unsigned
- Date/Time functions: Date, DateAdd, Timestamp, TimestampAdd
- String functions: Ascii, Bin, Length, Upper, Lower, Reverse, Trim
- Null functions: IsNull, NVL
- Special handling: Count (`str | Term` to support "*")

**Why?**
PR #815 introduced overly restrictive type annotations (`str | Field`) that caused mypy errors for valid pypika code. The issue is that `Field` is not the base type in pypika's hierarchy—`Term` is. 

Functions accept any `Term` subclass through the `builder.wrap_constant()` method, including:
- `Field` (table columns)
- `ArithmeticExpression` (e.g., `table.price * table.quantity`)
- Nested `Function` calls (e.g., `fn.Coalesce(fn.Avg(...), 0)`)
- Other `Term` subclasses

The `str | Field` annotation rejected valid patterns like `fn.Sum(table.a + table.b)` because arithmetic operations return `ArithmeticExpression`, not `Field`. Using `Term` as the base type correctly reflects pypika's design and resolves all mypy compatibility issues while maintaining backward compatibility.

**References**
- Reverts problematic typing from #815
- All 1,023 existing tests pass
- Verified against documented usage in README.rst and test suites

### DIFF
--- a/pypika/functions.py
+++ b/pypika/functions.py
@@ -6,7 +6,7 @@ from __future__ import annotations
 
 from pypika import Field
 from pypika.enums import SqlTypes
-from pypika.terms import AggregateFunction, Function, LiteralValue, Star
+from pypika.terms import AggregateFunction, Function, LiteralValue, Star, Term
 from pypika.utils import builder
 
 __author__ = "Timothy Heys"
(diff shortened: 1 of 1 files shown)


AzisK:
Looks good to me. Thank you for you work @noambloom

AzisK (approved):

PR #866, opened 2026-02-02
typefix: Add explicit type annotations to table attribute for mypy compatibility

noambloom:
**What changed?**
Added explicit type annotations to the `table` attribute in `Field` and `JSON` classes.

**Why?**
Mypy reported unreachable code errors due to implicit typing of the `table` attribute.

**Problem:**
```python
class Field(Criterion, JSON):
    # table was implicitly typed as None
    
    @property
    def get_table_name(self) -> str | None:
        if self.table:
            # mypy error: else branch unreachable because
            # isinstance() didn't properly narrow type
            return self.table if isinstance(self.table, str) else self.table.get_table_name()
        return None

### DIFF
--- a/pypika/terms.py
+++ b/pypika/terms.py
@@ -656,7 +656,6 @@ class Field(Criterion, JSON):
     def nodes_(self) -> Iterator[NodeT]:
         yield self
         if self.table is not None:
-            assert isinstance(self.table, Selectable)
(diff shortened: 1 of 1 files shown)


noambloom:
@AzisK Looks good to merge and release a patch i.e. 0.51.1

AzisK:
It's released

PR #3519, opened 2026-03-29
Fix SQLite Query class docstring

FoxRLY:
- Fixed docstring for SQLite Query class to reference SQLite instead of Microsoft SQL Server
- Changed naming of SQLite connected classes from `SQLLite` to conventional `SQLite`

I don't know the reason why `SQLite` objects are called `SQLLite` - maybe it's something important or contributes to backwards compatability. Because of that second commit is seen as optional by me


AzisK:
am I right to say that this is a breaking change?

FoxRLY:
@AzisK guess it is. Maybe making these 2 changes as separate PRs would be better if breaking changes are not appreciated

AzisK:
Breaking changes are never appreciated by the users. However, sometimes it is the way. Would need to bump the major version then

FoxRLY:
Ok, I'll change the scope of the PR to only fix docstring so major version change is not required. PR with name changes will come later

